PR #16, opened 2020-03-30
Register hook

dfalbel:
This PR turned out to be more complex than expected.

Since hooks are arbitrary R functions, they must run in the main
thread, but `backward` can call them from other threads..  In order to make sure, 
we call `backward` in a different thread
see `cpp_backward` and leave the main thread free to execute the R
calbacks that are pushed to the `event_loop_thread` as `packaged_tasks` to the `tasks` global.

However, the R hooks can by themselves call `backward` again and torch
does not allow us to use a different thread. so while we are waiting for
the hook to finish, we allow to execute tasks in the thread
the hook has been called.


dfalbel:
\gpu

PR #34, opened 2020-06-03
Rename package back to 'torch'

javierluraschi:
Fix for https://github.com/mlverse/torch/issues/30.

Currently, `rcpp` does not support naming the library that gets generated, instead it uses `Package` from the `DESCRIPTION` file. So I think we are left with limited options:
1) Send a PR to `rcpp` to include a new `LibName` attribute in the `DESCRIPTION` field.
2) Rename the library ourselves.

This PR goes for (2) as a start, which happens to work fine except that it currently generates two libraries `torch.so` and `torchpkg.so` which increases the size of the package 2x. This is fine during development but we should remove the extra library from being generated when release to CRAN, see comment in [tools/renamelib.R](https://github.com/mlverse/torch/pull/34/files#diff-356a22515ff105ce4f2863f90a0f1613R6).


zerweck:
I was looking for the reasons this package is named torch now and not torchr anymore. 
FWIW, I think there is one major problem with this: Finding relevant content _only_ for the R version via online search is currently very hard. Basically every google hit steers me to pytorch. 
Even adding "R" or "mlverse" to the search is often not helping. 
Maybe having a distinct Stackoverflow tag could help, but currently no one can assume the R connection just by the name. And, contrary to the keras and tensorflow packages being also named the same in R, these are implemented via reticulate so finding python examples is still useful as they can be easily ported. This is not the case here, so a distinct way of researching posts by other users would be a great benefit in my opinion.

dfalbel:
Hi @zerweck 

I can see the point that {torch} may not be the easiest way to find but, IMHO at the moment we just don't have many post and articles about torch for R online. 

FWIW if I search "torch for R" I don't see any PyTorch content at all and I think I see all the content that I am aware about the {torch} package.

zerweck:
Thank you Daniel for your response. 

Let me give an example of why I felt the need to post this:
The value in this package compared to https://github.com/f0nzie/rTorch (a reticulate implementation, like https://github.com/rstudio/tensorflow) is (to me, at least) to utilize R libraries that don't exist in the python world for the components such as the autograd features or custom loss functions.
If, in the future, someone shares a good example implementation utilizing R Code, it would help other users if it is easily found via Stack Overflow or Medium. I feel like a clear name might help users to identify the unique selling point of _this_ library compared to others.

Also, it might be worth to think about that torch used to be the name of the discontinued PyTorch predecessor Lua library https://en.wikipedia.org/wiki/Torch_(machine_learning) and could be confused. All other libraries from the current *torch ecosystem have a clear naming scheme: libtorch for the C++ backend, PyTorch for Python, Hasktorch for Haskell, etc.
rTorch and this library both exist for a specific need and it could be benefitial for users if both had a clear name stating their mission.

It seems as if RStudio does not have the same concerns as can be seen by naming the R implementations of tensorflow and keras the same as the python libraries they are calling. I just feel in these cases it's less of a problem due to their direct calling of the python code. But in the end, this is your decision and I just wanted to state my personal opinion on the matter.

PR #38, opened 2020-06-11
Faster class system

dfalbel:
This PR adds a class system, very similar to R6, to be used by the Tensor objects.

We use R7 to avoid the large times of R6 objects initialisation (`Tensor$new`).

Currently:

- Does not provide private fields (only methods)
- It doesn't support inheritance
- Doesn't check for duplicated methods, etc.


wlandau:
Interesting. I tried to do something similar at one point (using static code analysis on the constructor for inheritance) but ran into [memory issues](https://stackoverflow.com/questions/62513672/a-more-memory-efficient-way-to-mimick-message-passing-oop-in-r) because the methods were duplicated in memory (just like R6 and reference classes) and so memory consumption increased linearly with the number of objects. Happen to have any tips on lowering the memory footprint and still have a message-passing-like feel?

dfalbel:
The case for tensors is simpler because the only state we need to keep in the R side is the `ptr` object and all methods are static and depend only on this pointer. However R6 was re-creating all methods for each new tensor, which would be fine if we had few methods, but we have ~200. 
 
What I did here is to make sure that all methods are saved in a [common class environment ](https://github.com/mlverse/torch/pull/38/files#diff-e12220f4da2a1d0b1bb31b34b961aac0R17-R18) and then [overloaded the the `$` operator](https://github.com/mlverse/torch/pull/38/files#diff-e12220f4da2a1d0b1bb31b34b961aac0R55-R76) to look for methods in this environment.

PR #51, opened 2020-06-16
Update nn-rnn.R

jwijffels:


dfalbel:
Thanks!

jwijffels:
Note also that device=input$device() is not a function part of the object set by nn_rnn at line https://github.com/mlverse/torch/blob/master/R/nn-rnn.R#L136

dfalbel:
Opened #52 to track it

PR #74, opened 2020-06-19
Auto-install GPU version when CUDA detected and avoid prompts

javierluraschi:
- Auto-install GPU version when CUDA detected.
- Avoid installation prompts. There is precedent where binaries are downloaded when a package is built in Windows. 99% of the users will want to install Torch when the package loads, so we would be doing them a disservice if we prompt all the time. CRAN does not allow installing anything outside the package folder, but in this case, Torch and Lantern are being installed in the package folder, so it should be all good! Hopefully!


javierluraschi:
The OS X and Linux errors are unrelated, tests have passed in other runs.

PR #86, opened 2020-06-22
Enhance datasets API

### LINKED ISSUE #54: Think about the Datasets API
#48 implemented basic datasets support. In the end we have something like this:

```
x <- torch_randn(1000, 100)
y <- torch_randn(1000, 1)
dataset <- utils_dataset_tensor(x, y)
  
dl <- DataLoader$new(dataset = dataset, batch_size = 32)
 
iter <- dl$.iter()
b <- iter$.next()
```

In python,  you can easily iterate over a class that has an `__iter__` method, so the API looks like just:

```
for x, y in dl:
  ...
```

We need to think how this API would look like in R.

dfalbel:
In this PR we added a couple of functions to ease iteration of datasets. See examples below:

```
x <- torch_randn(100, 100)
y <- torch_randn(100, 1)
dataset <- utils_dataset_tensor(x, y)
dl <- utils_dataloader(dataset = dataset, batch_size = 32)

# iterating with a while loop
iter <- dataloader_make_iter(dl)
while(!is.null(batch <- dataloader_get_next(iter))) {
   expect_tensor(batch[[1]])
   expect_tensor(batch[[2]])
}
  
# iterating with an enum
for (batch in dataloader_enum(dl)) {
  expect_tensor(batch[[1]])
  expect_tensor(batch[[2]])    
}
```

Thiis fixes #54


dfalbel:
Changed to:

```
x <- torch_randn(100, 100)
y <- torch_randn(100, 1)
ds <- tensor_dataset(x, y)
dl <- dataloader(dataset = ds, batch_size = 32)

# iterating with a while loop
iter <- dataloader_make_iter(dl)
while(!is.null(batch <- dataloader_get_next(iter))) {
  expect_tensor(batch[[1]])
  expect_tensor(batch[[2]])
}

# iterating with an enum
for (batch in dataloader_enum(dl)) {
  expect_tensor(batch[[1]])
  expect_tensor(batch[[2]])    
}
```

Think for `dataloader_enum` yet.

javierluraschi:
I would do `dataloader_next()` instead of `dataloader_get_next()`. I think `dataloader()` is looking good now, so +1 from me!

DESCRIPTION
```diff
@@ -32,7 +32,9 @@ Suggests:
     covr,
     knitr,
     rmarkdown,
-    bit64
+    bit64,
+    readr,
```
javierluraschi, line 36: I don't think we should depend on `readr` here, can we use `read.csv`? There are a lot of dependencies in `readr`, so I much rather avoid the build times and breaking changes if possible. `fs` aand `bit64` make sense, but I'd really try to avoid `readr`.

PR #95, opened 2020-06-24
WIP: Get GPU lantern builds and installation working

javierluraschi:


javierluraschi:
Merging, might break installs temporarily while the GA rebuilds the libraries...

PR #115, opened 2020-06-25
Feature/stride

dfalbel:
Allow calling `t$stride()` with no arguments, like in python.


skeydan:
thanks (no review coz merged :-))

PR #121, opened 2020-06-25
WIP: Support for handling errors in Windows

javierluraschi:
Windows makes use of a different compiler MSVC with incompatible exception mechanisms than MinGW; therefore, we have to catch and check for errors ourselves. Fixes https://github.com/mlverse/torch/issues/31.


javierluraschi:
Some builds are going to fail since lantern has changed, but the Linux build should pass since lantern is rebuilt there.

PR #129, opened 2020-06-27
Enable windows exception test

javierluraschi:
Windows test should pass now...


javierluraschi:
Tested the failed autograd test in Windows with CMAKE, looks good, so merge should succeed after binaries get updated.

PR #154, opened 2020-07-01
add vignette for custom datasets

skeydan:
this was supposed to come nicely working ;-)))

But without "doing anything" ;-))) (as users do; meaning: no pulls, ..) `dataset()` stopped being found, and before that `prepare_penguin_data` wasn't found (((it all worked up to some point, apart from the training loop which I didn't get to test yet due to that sudden namespace thing)))

Perhaps you could just run it on your side and see if it works - at least until the training loop, where I might still hav to fix something? (I was suspecting that `purrr` inside the dataset was the culprit, but I definitely was already using `dplyr` before, and now it fails for me even without the call to `purrr`...)

I should perhaps add that I'm still using the `devtools::lload_all` workflow, not `install` (which idk if it should work by now)

### DIFF
--- a/vignettes/loading-data.Rmd
+++ b/vignettes/loading-data.Rmd
@@ -2,7 +2,7 @@
 title: "Loading data"
 output: rmarkdown::html_vignette
 vignette: >
-  %\VignetteIndexEntry{Creating tensors}
+  %\VignetteIndexEntry{Loading data}
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
@@ -15,7 +15,7 @@ knitr::opts_chunk$set(
 ```
 
 ```{r setup}
-#library(torch)
+library(torch)
 ```
 
 ## Datasets and data loaders
@@ -59,10 +59,7 @@ your own `dataset` to train on [Allison Horst\'s penguins](https://github.com/al
 ## A custom dataset
 
 ```{r}
-# remotes::install_github("allisonhorst/palmerpenguins")
 library(palmerpenguins)
-library(dplyr)
-library(purrr)
 
 penguins
 ```
@@ -103,11 +100,14 @@ penguins_dataset <- dataset(
   },
   
   prepare_penguin_data = function() {
+    
+    input <- na.omit(penguins) 
     # conveniently, the categorical data are already factors
-    input <- na.omit(penguins) %>%
-      mutate_if(is.factor, compose(partial(`-`, ... = , 1), as.integer)) %>%
-      # mutate_if(is.factor, as.integer) %>%
-      as.matrix()
+    input$species <- as.numeric(input$species) - 1
+    input$island <- as.numeric(input$island) - 1
+    input$sex <- as.numeric(input$sex) - 1
+    
+    input <- as.matrix(input)
     torch_tensor(input)
   }
 )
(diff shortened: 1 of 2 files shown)


vignettes/loading-data.Rmd
```diff
@@ -0,0 +1,198 @@
+---
+title: "Loading data"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Creating tensors}
```
dfalbel, line 5: We need to change the title here too

vignettes/loading-data.Rmd
```diff
@@ -0,0 +1,198 @@
...
+
+While this may sound complicated, it is not at all. The base logic is straightforward -- complexity will, naturally, correlate
+with how involved your preprocessing is. To provide you with a simple but functional prototype, here we show how to create
+your own `dataset` to train on [Allison Horst\'s penguins](https://github.com/allisonhorst/palmerpenguins).
+
+## A custom dataset
+
+```{r}
+# remotes::install_github("allisonhorst/palmerpenguins")
+library(palmerpenguins)
```
dfalbel, line 63: I think we need to add palmerpenguins to the Suggests in the DESCRIPTION file
If it's only on github we also need the Remotes field

Same for `dplyr` and `purrr`. (Perhaps it's better to avoid them to not increase the CI testing time)

vignettes/loading-data.Rmd
```diff
@@ -0,0 +1,198 @@
...
+  
+  .length = function() {
+    self$data$size()[[1]]
+  },
+  
+  prepare_penguin_data = function() {
+    # conveniently, the categorical data are already factors
+    input <- na.omit(penguins) %>%
+      mutate_if(is.factor, compose(partial(`-`, ... = , 1), as.integer)) %>%
+      # mutate_if(is.factor, as.integer) %>%
```
dfalbel, line 109: I think it's more readable to do: 

```
mutate_if(is.factor, ~as.integer(x) -1)
```

But not a strong opinion :)

vignettes/loading-data.Rmd
skeydan: hehe right ;-)

vignettes/loading-data.Rmd
skeydan: yeah, I see... removing `purrr` is also what you suggest below  - and removing `dplyr` isn't that hard either so I'll do that :-)

skeydan:
happy you like it! I had to figure out how to use it anyway, so why not create a vignette at the same time ... :-)

vignettes/loading-data.Rmd
skeydan: hm, for palmerpenguins ... this: https://stackoverflow.com/questions/36105257/how-to-make-r-package-recommend-a-package-hosted-on-github sounds pretty awkward ... and it seems like they're in the process of putting it on CRAN ... https://github.com/allisonhorst/palmerpenguins/commit/8e1061dc0b3d0cbb53bea0545f651950b2109dca

perhaps i just add it to Suggests for now?

skeydan:
just pushed .... strangely, i still get 
> Error in prepare_penguin_data() : 
  could not find function "prepare_penguin_data"

... i'll retry later, after rebuilding for all the cool new stuff you added :-)

... does it run for you?

dfalbel:
Fixed! there was a `self$` missing

skeydan:
oh, cool! thanks!

(((i think the reason it ever ran must have been that i had it defined globally at first, for testing ;-)))

PR #166, opened 2020-07-03
Follow ups on Windows tests

javierluraschi:
Trying to enable a few more tests in Windows...


javierluraschi:
A few more tests got fixed with this, osx and windows use pre-built lantern so they are expected to fail in PRs.

PR #171, opened 2020-07-04
Resolves #164: Implement more R generic methods

### LINKED ISSUE #164: Allow more generic R methods
Hi, thanks for the effort in developing this package 👍 . I've been following the development of the package and am quite excited about the idea. After quickly trying the tutorials, I notice that `operators.R` currently only implements the S3 methods for some operations. I wonder if there will be any plans to implement more  math generic methods such as `log`, `sum`, `mean`, etc. In my case, this would allow me to reuse my existing R code base and apply autograd easily. If this sounds interesting but isn't the current focus, I'm keen to contribute 😃

qykong:
This pull request implements the following R generic functions along with the corresponding test cases:
> %%, %/%, &, |, !, abs, sign, sqrt, ceiling, floor, trunc, cumsum, log, log10, log2, log1p, exp, expm1, cos, cosh, sin, sinh, tan, tanh, acos, asin, atan, max, min, prod, sum

While most functions are straightforward to implement, several special cases are noted here:

- `log`: default R has a `base` argument but I didn't find the corresponding in pytorch, so I did a trick as`torch_log(x)/torch_log(base)`
- `sum` and `prod`: `nan`s in the tensor will be ignored if `na.rm=T`. If single tensor is given, one can specify `dim` and `keepdim`. If multiple tensors are given, `dim` argument is not allowed and only a single value will be computed from them.
- `min` and `max`: `nan`s in the tensor will be ignored if `na.rm=T`. Only a single value is returned.

### DIFF
--- a/R/operators.R
+++ b/R/operators.R
@@ -30,6 +30,16 @@
 }
 
 #' @export
+`%%.torch_tensor` <- function(e1, e2) {
+  torch_fmod(e1, e2)
+}
+
+#' @export
+`%/%.torch_tensor` <- function(e1, e2) {
+  torch_floor_divide(e1, e2)
+}
+
+#' @export
 `>=.torch_tensor` <- function(e1, e2) {
   torch_ge(e1, e2)
 }
@@ -57,6 +67,21 @@
 #' @export
 `!=.torch_tensor` <- function(e1, e2) {
   torch_ne(e1, e2)
+}
+
+#' @export
+`&.torch_tensor` <- function(e1, e2) {
+  torch_logical_and(e1, e2)
+}
+
+#' @export
+`|.torch_tensor` <- function(e1, e2) {
+  torch_logical_or(e1, e2)
+}
+
+#' @export
+`!.torch_tensor` <- function(x) {
+  torch_logical_not(x)
 }
 
 #' @export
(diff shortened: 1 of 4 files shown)


R/operators.R
```diff
@@ -89,7 +114,167 @@ as.double.torch_tensor <- function(x, ...) {
...
+# helper function to replace nan in tensors
+replace_nan <- function(tensor, value) {
+  torch_where(torch_isnan(tensor),
+              torch_full_like(tensor, value),
+              tensor)
+}
+
+#' @export
+max.torch_tensor <- function(..., na.rm = FALSE) {
+  l <- if (na.rm) lapply(list(...), function(tensor) replace_nan(tensor, -Inf)) else list(...)
```
dfalbel, line 240: I would ignore the `na.rm` argument  here - perhaps raising an error if it's not `FALSE`. See my comment in  #172 
Same for all other functions that have  an `na.rm` argument.

dfalbel:
Thanks @qykong ! It looks great!

R/operators.R
qykong: sounds good, I'll fix this and push shortly

qykong: Updated and pushed.

dfalbel:
We loose precision when converting R objects to tensors because we default to float32 while R uses float64. We should check with a higher tolerance.

From the CI:

```
sum works: failure: as.numeric(sum(torch_tensor(x))) not equal to sum(x).
1/1 mismatches
[1] 54.5 - 54.5 == 7.85e-06 
  callstack:
      
prod works: 0.00670194625854492
sum works: failure: as.numeric(sum(torch_tensor(x), torch_tensor(y))) not equal to sum(x, y).
1/1 mismatches
[1] 60.5 - 60.5 == 7.85e-06 
  callstack:
```

qykong:
hi @dfalbel , i just noticed a bug in handling cases like `-x`. For example, `- torch_tensor(1)` gives `1` rather than `-1`. As this is a quick fix in the same file, I did that and added a test case within this pull request.

dfalbel:
Awesome! thanks very much @qykong !

PR #187, opened 2020-07-08
Prepare lantern for long-tail of error handling

javierluraschi:
Appart from autograd and autogen functions, we also need to handle errors across all our custom lantern functions. This PR adds support for a `LANTERN_HOST_HANDLER` define hook, to allow torch to define a custom try/catch across all functions without having to add `LANTERN_ERROR_HANDLES` everywhere in the code.


javierluraschi:
The non-ubuntu tests are expected to fail since lantern is not being built, will follow up if they break in master.

Issue #149, opened 2020-06-30
as_array convertion for long type


dfalbel:
Should make the following work

```
as_array(torch_tensor(1, dtype = torch_long()))
```

PR #225, opened 2020-08-19
default to long

skeydan:


dfalbel:
Two tests are failing with this change. I think it's the expectation that is wrong now. Do you mind fixing it?

skeydan:
already on it :-) (((may have questions though)))

skeydan:
I think that for
``` 
x <- 1:4
expect_equal_to_r(torch_tensor(x), x)
```

I need to add int64 stuff in lantern everywhere, analogous to this:

image

right?

dfalbel:
I would just make something like:

```
expect_equal_to_r(torch_tensor(x)$to(dtype = torch_int()), x)
```

for now. Adding support for converting back from torch_long() is trickier to implement. See also 
#149

skeydan:
done :-)
(hopefully i caught them all, because running `devtools::test()` locally segfaults for me...)

totally forgot about that issue #149...

dfalbel:
Thanks! Working now :D

PR #249, opened 2020-09-17
Fix ordering for parallel make

dfalbel:
declare rename dependency on $(SHLIB) so it's always run in the correct order.


javierluraschi:
Looks great! Thanks @kevinushey for spotting this one!

PR #284, opened 2020-10-06
Adadelta optimizer

krzjoa:
Adadelta optimizer added


dfalbel:
That's really awesome @krzjoa ! Thanks very very much!

dfalbel (approved):

PR #291, opened 2020-10-12
allow for cuda 11.0

skeydan:


skeydan:
On linux this should work - let's see what CI says for Win...

(It's of course not a nice workaround, but I don't see a better one until PT provide official downloads...)

dfalbel:
We don't have CI tests for CUDA 11.0 so we won't know if it works or not on Windows. Wouldn't it be better to recommend installing CUDA 10.2? It's possible to have two coexisting versions of the toolkit in a system.

skeydan:
like we just talked about, removed windows

ekianjo:
Thanks for the fix!

skeydan:
Sorry but I'm afraid we'll have to either revert this or see how we could circumvent this:
```
Warning message:
Failed to install Torch, manually run install_torch(). /home/skeydan/R/x86_64-pc-linux-gnu-library/3.6/torch/deps/liblantern.so - libcudart.so.10.2: cannot open shared object file: No such file or directory 
```

Unfortunately it looks like this only works if you compile lantern yourself, like I do on my laptop, but once you use `remotes::install_github` it wants to compile against cuda 10.2 :-(

I'll add code to revert this to the (unrelated) PR https://github.com/mlverse/torch/pull/293, so I can install remotely from that PR...

dfalbel:
The solution then is to add CUDA 11 installation steps here:

https://github.com/mlverse/torch/blob/master/.github/workflows/lantern.yaml#L133-L148

And also in the CMakeLists.txt file 

https://github.com/mlverse/torch/blob/master/lantern/CMakeLists.txt#L53

PR #297, opened 2020-10-13
Rprop optimizer

krzjoa:


dfalbel:
Thanks very much @krzjoa !!

PR #309, opened 2020-10-19
Sparsemax

dfalbel:
Added an implementation of sparsemax.


dfalbel:
Passing on linux is sufficient for merging as we need to rebuild lantern binaries.

Issue #240, opened 2020-09-14
missing tensor methods


dfalbel:
- bool
- element_size

dirkschumacher:
`real` is only supported for complex types. So those would need to be implemented first.

PR #322, opened 2020-10-24
Implement element_size

dirkschumacher:
Hi @dfalbel . It still crashes on my machine, but I thought this small PR would be a good way to discuss what I am missing and to learn how to contribute efficiently.

```R
callr::r(function() { 
  library(torch)
  x <- torch_tensor(1)
  result <- x$element_size()
}) # crashes R session
```

Relates to #240

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -7105,6 +7105,10 @@ cpp_as_array <- function(x) {
     .Call('_torch_cpp_as_array', PACKAGE = 'torchpkg', x)
 }
 
+cpp_tensor_element_size <- function(x) {
+    .Call('_torch_cpp_tensor_element_size', PACKAGE = 'torchpkg', x)
+}
+
 cpp_tensor_dim <- function(x) {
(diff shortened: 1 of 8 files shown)


dfalbel:
Hi @dirkschumacher 

Thanks very much! 

Everything looks correct except the missing entry here:

https://github.com/mlverse/torch/blob/2bc73cba2e5aeb8118374ef0acf21c9e903f8d6f/lantern/include/lantern/lantern.h#L4309

We need this do dynamically load all symbols from lantern.

dirkschumacher:
Hmm. Whats going on with Windows?

Would be one improvement to check at runtime that `torch_tensor` does have a correct `dtype` (i.e. no quantized datatype)?

dfalbel:
The GPU worker is down at the moment... we can merge this PR now.
We could have the runtime check for the correct dtype, but I don't think it's necessary. Currently we get the same error as in PyTorch. 
The quantization API is still experimental, so they might improve the error message LibTorch side in the future.

PR #323, opened 2020-10-24
Add tensor$bool

dirkschumacher:
According to the pytorch docs:
> self.bool() is equivalent to self.to(torch.bool)

I haven't found a way to check the memory format, so I just test for no errors, assuming that `$to` works correctly.

Relates to #240

### DIFF
--- a/tests/testthat/test-tensor.R
+++ b/tests/testthat/test-tensor.R
@@ -309,10 +309,10 @@ test_that("tensor$bool works", {
   x <- torch_tensor(c(1,0,1))
   result <- x$bool()
   expected <- x$to(torch_bool())
-  expect_equal(result, expected)
+  expect_equal_to_tensor(result, expected)
 
   expect_silent(
     result <- x$bool(memory_format = torch_contiguous_format())
   )
-  expect_equal(result, expected)
+  expect_equal_to_tensor(result, expected)
 })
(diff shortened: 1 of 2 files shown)


tests/testthat/test-tensor.R
```diff
@@ -304,3 +304,15 @@ test_that("element_size works", {
     expect_true(length(result) == 1)
   }
 })
+
+test_that("tensor$bool works", {
+  x <- torch_tensor(c(1,0,1))
+  result <- x$bool()
+  expected <- x$to(torch_bool())
+  expect_equal(result, expected)
```
dfalbel, line 312: I am not sure what `expect_equal` does, but it doesn't test equality of tensors. We are using `expect_equal_to_tensor(x, y)` to compare tensors in the tests.

dirkschumacher: Changed it. Thanks

dfalbel:
Usually memory formats will only make sense for 4d tensors, eg:

``` r
library(torch)
x <- torch_randn(10, 10, 10, 10)
x$stride()
#> [1] 1000  100   10    1
x$to(memory_format = torch_channels_last_format())$stride()
#> [1] 1000    1  100   10
```

I will add a test for this later in `to`.

dfalbel:
Thanks @dirkschumacher ! LGTM! I'll merge when tests passes

Issue #324, opened 2020-10-25
Check hashes of downloaded libraries


dirkschumacher:
One option to increase securiy of the package is to hardcode the hashes of the files downloaded in [`R/install.R`](https://github.com/mlverse/torch/blob/master/R/install.R) and check the hashes after downloading.

This protects against corrupted files during transmission and also protects somewhat againts attacks on the storage servers or MITM attacks.

Just an idea.

dfalbel:
I agree! I don't see a way of doing it for the development version of Lantern, as the hashes will change all the time.
But we can definitely do it for LibTorch binaries and for Lantern in the CRAN releases.

dirkschumacher:
We could use `md5` hashes checked with `tools::md5sum`, not the best, but those require no additional packages.

PR #325, opened 2020-10-25
Check md5 hashes of pytorch libraries after download

dirkschumacher:
A first proposal to check the pytorch files after installation. I am open to any suggestions.

Current idea is to hardcode md5 hashes along the URLs to pytorch in the `install_config`. Then during installation compare the two hashes and give a somewhat informative error messages, if that fails.
The downloaded file now also gets deleted after the function is exited. In case it was really a malicious file it would be nice to get rid of it asap :)

Relates to #324

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,8 @@ Imports:
     utils,
     stats,
     bit64,
-    magrittr
+    magrittr,
+    tools
 RoxygenNote: 7.1.1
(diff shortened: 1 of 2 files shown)


dfalbel:
LGTM! I think for lantern we can have `NULL` by default during development and add the hashes before each CRAN release.
Also it's nice to delete the file as soon as possible.

PR #327, opened 2020-10-25
[doc] update getting-started

hbaniecki:
text/code mismatch in https://torch.mlverse.org/docs/articles/getting-started/control-flow-and-weight-sharing.html
`The first argument to the Adam constructor tells the [...]` 

new result:

```
Step: 1 : 1.01752 
Step: 100 : 0.6261083 
Step: 200 : 0.2703869 
Step: 300 : 0.1696484 
Step: 400 : 0.06325339 
Step: 500 : 0.03317686 
```


dfalbel:
Thanks very much @hbaniecki !

PR #328, opened 2020-10-25
Refactor most of the optimizer step functions.

dirkschumacher:
I noticed a bit of duplicated code. The common part is now extracted into a private helper function.

### DIFF
--- a/R/optim-adadelta.R
+++ b/R/optim-adadelta.R
@@ -29,63 +29,38 @@ optim_Adadelta <- R6::R6Class(
     },
     
     step = function(closure = NULL){
-      
-      with_no_grad({
-        
-        loss <- NULL
-        
-        if (!is.null(closure)) {
-          with_enable_grad({
-            loss <- closure()
-          })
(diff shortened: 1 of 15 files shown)


dfalbel:
Thanks @dirkschumacher ! This looks great!  I'll review this PR more carefully tomorrow, my main concern is if:

https://github.com/mlverse/torch/pull/328/commits/129063f4398a46318a7a0abbc05845f2491a04db#diff-4de20d78bee9258794b8ce93bff5f31cd1e99869faa75d089fe46b8f0ddb39a2R41-R42

would  change `param` only in the function context of if it would also modify the reference inside `self` and I am not sure the tests would capture that.

dirkschumacher:
Yeah I was thinking something similiar. What is the intented behaviour here? Modification of `self`or not?

dfalbel:
It's expected that `self` is modified.

dfalbel (approved):
I added a test to check that the state is correctly modified. So we are good to go! Thanks very much @dirkschumacher !

Issue #336, opened 2020-10-29
use nvcc -- version instead of nvidia-smi to detect installed CUDA version (Linux)


strimmerlab:
On Linux 
https://github.com/mlverse/torch/blob/master/R/install.R
uses the command "nvidia-smi" to detect the installed CUDA version.  However this returns the highest CUDA version that the installed graphics driver is capable to support, not the actually installed version of CUDA!  The installed version is obtained by running the command "nvcc --version":

The CUDA version in Debian 10 is 10.1:
https://packages.debian.org/buster-backports/nvidia-cuda-toolkit

This gives the correct version:
> nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2019 NVIDIA Corporation
Built on Sun_Jul_28_19:07:16_PDT_2019
Cuda compilation tools, release 10.1, V10.1.243

This gives the wrong version (it shows the CUDA version the driver version 450.66 is able to support)
> nvidia-smi
...    
| NVIDIA-SMI 450.66       Driver Version: 450.66       CUDA Version: 11.0     |
....

See also:
https://stackoverflow.com/questions/53422407/different-cuda-versions-shown-by-nvcc-and-nvidia-smi

dfalbel:
Thanks for the report! Will fix!

strimmerlab:
Thank you!

PR #363, opened 2020-11-13
use nvcc to get compiler version

skeydan:
https://github.com/mlverse/torch/issues/336


skeydan:
Hm... this is not as easy... the system I'm now trying it on has `nvcc` version 10.0, which does not appear as an allowable CUDA version here: https://pytorch.org/ ... which is why we're not allowing it ...

PR #377, opened 2020-11-24
Start experimenting with jit support

dfalbel:
The scope of this PR is to implement the equivalent to trace an R function to a ScriptFunction(https://github.com/pytorch/pytorch/blob/fa560ceb9c8cd0b7e40c3e7ec7808c01d2399377/torch/csrc/jit/python/script_init.cpp#L1111-L1196) in PyTorch.

- We will only add tracing, ie. loops that depend on data attributes are not supported.
- The script function doesn't act like an nn_module, meaning that you don't train it. However we will be able to save it to a file a load it in python or c++...

Support for ScriptModules is possible but requires implementing a ModuleWrapper (https://github.com/pytorch/pytorch/blob/fa560ceb9c8cd0b7e40c3e7ec7808c01d2399377/torch/nn/cpp.py#L49)  

TODOS:

- [x] Implement `torch::jit::Stack` <-> R lists. (very basic support implemented)
- [x] Convert the R function to a function taking and returning stacks. (done in the c++ side only)
- [x] Figure out what to return from the traceable fn? The graph or the function created in the compilation unit?
- [x] R functionality to easily create R function taking and returning stacks
- [x] R capsule for the pointer returned by the trace_fn, with possibility to print the graph and etc. Similar to the ScriptFunction
- [x] JIT save and JIT load
- [x] Allow printing the generated graph
- [x] Add tests for more complex functions
- [x] Complete support for most IValues when creating Stacks.
- [x] Allow loading JIT functions (created in python or R)
- [x] Dispose of memory allocated when creating the traceable function.


skeydan:
hi @dfalbel , awesome!

Just to check I understand correctly, this means we'll have support for what is described here:

https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html#tracing-modules

- tracing, as opposed to scripting - , and we assume that you don't want to train a scripted model, but do the tracing call after training, such that the final parameters are used?

Thanks!

dfalbel:
Yes exactly! The main idea is to be able to easily export models to be run out of R.
In the future we can support tracing modules for training too (which would give some speedups).

skeydan:
awesome!

PR #386, opened 2020-12-08
Add build env guide for Torch RStduio on Docker

y-vectorfield:
Hello, @dfalbel 
I added environmental construction guide for Torch RStudio with CUDA on Docker. We can build and use a docker container which is included RStudio, Torch for R(with CUDA) and so on.


dfalbel (approved):
Looks fantastic! Thanks very much @y-vectorfield

PR #402, opened 2020-12-19
Custom CPU allocator

dfalbel:
This PR implements a custom CPU allocator in Lantern so we can smartly call the R `gc` whenever we need more memory.
Currently this is just a dummy allocator that just prints how much memory it has allocated.


emauryg:
could this also be implemented for GPU?

dfalbel:
It has been implemented for the GPU here: https://github.com/mlverse/torch/pull/456

PR #411, opened 2020-12-26
Use `nzchar` to check non-empty strings

hsbadr:
`nzchar` is a fast way to find out if elements of a character vector are non-empty strings.


dfalbel:
Thanks very much!

hsbadr:
@dfalbel Thank you! I'm glad that you allowed R community to use Torch natively from R. Looking forward to making it stable.

dfalbel:
Yeah looking forward to making it stable too! We still need some iterations but hopefully will get there :)

PR #438, opened 2021-01-16
Feature/install torch timeout as a fix to #437

### LINKED ISSUE #437: Please include a higher `timeout=` option in `install_torch()` to prevent download file 60s timeout
Hello, 

Initial `library(torch)` or manual `install_torch()` very often fails due to large file download timeout, as mentionned by https://github.com/mlverse/torch/issues/186#issuecomment-735576520

### Current situation
Usual home internet connection do not succeed to download libtorch (130MB+ ) before the default `timeout=60`, making the first contact with the package frustrating, and diverting the user from the goal to build and train a neural network.

Even if problem is made explicit during the `install_torch()` process by the log : 
```(r)
Warning messages:
1: In utils::download.file(library_url, temp_file) :
  downloaded length 44901568 != reported length 141774525
2: In utils::download.file(library_url, temp_file) :
  URL 'https://download.pytorch.org/libtorch/cpu/libtorch-macos-1.7.1.zip': Timeout of 60 seconds was reached
3: Failed to install Torch, manually run install_torch(). download from 'https://download.pytorch.org/libtorch/cpu/libtorch-macos-1.7.1.zip' failed 
```
But it is not obvious for most users on how to increase the timeout with the global `options(timeout = 360)`

## Proposed improvement
add a `timeout=360` option in the `install_torch()` function so that
- solution to the problem is self contained in the documentation
- an adapted default value gives a much higher success-rate of initial installation 

Thanks a lot

cregouby:
This is a proposal to fix #437 
Hope it helps.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -174,7 +174,7 @@ install_os <- function() {
   tolower(Sys.info()[["sysname"]])
 }
 
-lantern_install_libs <- function(version, type, install_path) {
+lantern_install_libs <- function(version, type, install_path, install_config) {
   current_os <- install_os()
   
   if (!version %in% names(install_config))
@@ -290,7 +290,8 @@ install_type <- function(version) {
 #' @param type The installation type for Torch. Valid values are \code{"cpu"} or the 'CUDA' version.
 #' @param reinstall Re-install Torch even if its already installed?
 #' @param path Optional path to install or check for an already existing installation.
-#' @param ... other optional arguments (like `load` for manual installation.)
+#' @param ... other optional arguments (like \code{`load=FALSE`} for manual installation, \code{`timeout=240`} for
+#' large file downloads.)
 #' 
 #' @details 
 #' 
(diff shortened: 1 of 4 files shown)


R/install.R
```diff
@@ -312,9 +315,14 @@ install_torch <- function(version = "1.7.1", type = install_type(version = versi
     dir.create(path, showWarnings = FALSE)
   }
   
+  saved_timeout <- getOption('timeout')
+  options(timeout = timeout)
```
dfalbel, line 319: Could we use `withr::with_options(list(timeout = ...)` instead? We might not get into the `options(timeout = saved_timeout)` call if the function errors in between.

dfalbel (changes requested):
Thanks @cregouby ! Looks good! I just added a comment.

R/install.R
cregouby: Sure ! I'll do it, thanks for the review .

PR #462, opened 2021-02-04
Normal distribution

krzjoa:
I'm not sure if you have any plans how to prefix **distibutions**, but for now I wrote the first distribution function using *distr_* at the beginning. I can modify it as you wish.


tests/testthat/test-distribution-normal.R
```diff
@@ -0,0 +1,45 @@
...
+  scale       <- torch_randn(5, 5)$abs()$requires_grad_()
+  loc_1d      <- torch_randn(1, requires_grad=TRUE)
+  scale_1d    <- torch_randn(1)$abs()$requires_grad_()
+  
+  expect_equal(distr_normal(loc, scale)$sample()$size(), c(5, 5))
+  expect_equal(distr_normal(loc, scale)$sample(7)$size(), c(7, 5, 5))
+  expect_equal(distr_normal(loc_1d, scale_1d)$sample(1)$size(), c(1, 1))
+  expect_equal(distr_normal(loc_1d, scale_1d)$sample()$size(), 1)
+  # expect_equal(distr_normal(0.2, .6)$sample(1)$size(), 1)
+  # expect_equal(distr_normal(-0.7, 50.0)$sample()$size(), 1)
```
dfalbel, line 17: I'd add tests making sure the constraints are used and respected. A single `expect_error` would be enough.

tests/testthat/test-distribution-normal.R
```diff
@@ -0,0 +1,45 @@
...
+  loc         <- torch_randn(5, 5, requires_grad=TRUE)
+  scale       <- torch_randn(5, 5)$abs()$requires_grad_()
+  loc_1d      <- torch_randn(1, requires_grad=TRUE)
+  scale_1d    <- torch_randn(1)$abs()$requires_grad_()
+  
+  expect_equal(distr_normal(loc, scale)$sample()$size(), c(5, 5))
+  expect_equal(distr_normal(loc, scale)$sample(7)$size(), c(7, 5, 5))
+  expect_equal(distr_normal(loc_1d, scale_1d)$sample(1)$size(), c(1, 1))
+  expect_equal(distr_normal(loc_1d, scale_1d)$sample()$size(), 1)
+  # expect_equal(distr_normal(0.2, .6)$sample(1)$size(), 1)
```
dfalbel, line 16: Can we uncomment this tests? Seems like they should work given line 26 below.

dfalbel (approved):
Looks great! THanks very much @krzjoa ! 

I like `distr_`, `distribution_` seems like too long...
I just added some comments regarding the tests, but I am ready to merge if you think it's better to add more tests in other PR.

tests/testthat/test-distribution-normal.R
krzjoa: Yes, you're right. I'll add some tests.

PR #464, opened 2021-02-04
typo in auth.txt; ahth.txt to auth.txt and "pass word" to "password"

issactoast:
I found a typo on docker/build_env_guide.md line 38-39. I think what you intended was to put "rstudio" in auth.txt file.


dfalbel (approved):
Thanks! You are right!

Issue #473, opened 2021-02-15
nnf_group_norm causing R to crash


Athospd:
`nnf_group_norm()` is causing R to crash. On the code below, we expect `nnf_group_norm(tt, 3)` and `nnf_layer_norm(tt, c(3,3))` to return the same output.

``` r
library(torch)
tt <- torch_tensor(t(matrix(1:3, 3, 3)) * 1.)
nnf_group_norm(tt, 3) # <------ CRASHES (do not run)

nnf_layer_norm(tt, c(3,3))
#> torch_tensor
#> -1.2247  0.0000  1.2247
#> -1.2247  0.0000  1.2247
#> -1.2247  0.0000  1.2247
#> [ CPUFloatType{3,3} ]
```

<sup>Created on 2021-02-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Python version

![image](https://user-images.githubusercontent.com/1925102/108002830-90039780-6fcf-11eb-8bd7-48ebcfa4d920.png)

``` r
library(torch)
sessionInfo()
#> R version 4.0.2 (2020-06-22)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 20.04.1 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
#> LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0
#> 
#> locale:
#>  [1] LC_CTYPE=pt_BR.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=pt_BR.UTF-8        LC_COLLATE=en_US.UTF-8    
#>  [5] LC_MONETARY=pt_BR.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=pt_BR.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
#> [11] LC_MEASUREMENT=pt_BR.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] torch_0.2.1.9000
#> 
#> loaded via a namespace (and not attached):
[...]

PR #474, opened 2021-02-16
nn_group_norm

Athospd:
Implements nn_group_norm()

source:

[https://pytorch.org/docs/stable/_modules/torch/nn/modules/normalization.html#GroupNorm]()

PS: issue #473 provents test to succeed.


dfalbel:
Thanks very much @Athospd !
I also pushed a fix for #473 in this PR :)

PR #492, opened 2021-03-10
as.list.nn_module_list

kevinykuo:
Convenient for applying over the module list. We have `[[` and `length()` so might as well expose this?


dfalbel:
Thanks @kevinykuo !

PR #496, opened 2021-03-15
fix multihead attention

kevinykuo:
Don't think this was complete when it was exported since it doesn't run on either dev or CRAN. This change gets it kinda working for my current use case, but might be good to test more before release.


dfalbel:
Thanks @kevinykuo ! I'll merge and open an issue to track progress on other cases.

PR #506, opened 2021-03-26
make arange include end index

skeydan:


skeydan:
Only change is in

 https://github.com/mlverse/torch/blob/0555381cc1ff3fcbb76f36329811124daf18c920/R/creation-ops.R#L253,

everything else examples/tests ... When doing the latter, I noticed a few problems that will need handling:
(no. 2 and 3 just came "by chance" I think)


```
> #1 should not include 3
> torch_arange(1, 2.5, 0.5)
torch_tensor
 1.0000
 1.5000
 2.0000
 2.5000
 3.0000
[ CPUFloatType{5} ]


> #2 index should be 4
> x <- torch_arange(1, 6)
> torch_kthvalue(x, 4)
[[1]]
torch_tensor
4
[ CPUFloatType{} ]

[[2]]
torch_tensor
3
[ CPULongType{} ]


> #3 should throw error
> torch_arange(1,7)$resize_(c(2,3))
torch_tensor
 1  2  3
 4  5  6
[ CPUFloatType{2,3} ]

File: R/creation-ops.R [hunk 1]
```diff
@@ -250,7 +250,7 @@ torch_arange <- function(start, end, step = 1, dtype = NULL, layout = torch_stri
     )
   )
   args$start <- start
-  args$end <- end
+  args$end <- end + 1
```
dfalbel, line 253: Perhaps we should make `end - step` to fix the first error you mentioned?

skeydan:
you mean this right?

```
args$end <- end + 1 - step
```

that fixes the above, but then I get 

```
> torch_arange(0,1,1)
torch_tensor
 0
[ CPUFloatType{1} ]
```

which probably isn't what we want ...

dfalbel:
Maybe `end + step` then? does that make sense? didn't stop to think too much

skeydan:
Hm, by now I'm at

```
args$end <- if (end %% step == 0 && !(end + step == 0)) end + step else end
```

which at least satisfies a bunch of tests I've added (including more non-integer and/or negative steps):

```
x <- torch_arange(0, 2, 0.5)
  expect_equal(length(x), 5)
  
  x <- torch_arange(0, 2.5, 0.5)
  expect_equal(length(x), 6)
  
  x <- torch_arange(0, 2.5, 0.51)
  expect_equal(length(x), 5)
  
  x <- torch_arange(0, 2.5, 0.49)
  expect_equal(length(x), 6)

  x <- torch_arange(0, 1, 1)
  expect_equal(length(x), 2)
  
  x <- torch_arange(2.5, 1, -0.5)
  expect_equal(length(x), 4)
  
  x <- torch_arange(2.5, 1, -1)
  expect_equal(length(x), 2)
  
  x <- torch_arange(2.5, 1, -0.51)
  expect_equal(length(x), 3)
  
  x <- torch_arange(2.5, 1, -0.49)
  expect_equal(length(x), 4)
```

But this is really is an instance of "test-driven" development, so I can't say anything beyond "it works correctly for all tests I could think of" ...

It is probably better than not making the change at all though.

dfalbel:
you have to call `source("tools/document.R")` and everything should work fine!

dfalbel:
Thanks very much @skeydan !

PR #517, opened 2021-04-09
add strong_wolfe 

skeydan:


skeydan:
TBD:

- fix error on _lbfgs works with strong wolfe_
- remove debugging output
- remove either inline or global definition of `flower()`
- compare output with Python (`rosenbrock, flower)`

dfalbel:
Hey @skeydan,

I adjusted a few calls and it runs fine now, I didn't do any additional testing though... Do you want to add more tests?

skeydan:
Hi @dfalbel this should be ready to merge now :-)

I've added one more test that triggers another code path, and by now I've seen quite a few different paths (although I'm sure there's at least one that has not been entered).

For that reason (and others), I would like to leave in the (commented) debugging output, as it could potentially be very useful when analyzing issues.

In addition to the unit tests, I also did a more complex side-by-side comparison of Python vs. R, and obtained very similar results.

dfalbel:
Yeah, it seems unrelated :( Not sure what this is! 
Thanks @skeydan ! Looks great :)

PR #538, opened 2021-04-29
Try building a non-ABI version of liblantern

dfalbel:


hsbadr:
@dfalbel JFYI. I'm building my own `libtorch` and `liblantern` for because the prebuilt binaries cause segmentation fault on my system, mainly because I build R from source code using Intel MKL and compilers; I think this also improves the performance but I haven't run ay benchmarks yet. Are you willing to allow building `libtorch` from source code?

dfalbel:
Yeah, I think `TORCH_HOME` is the right option if you want to build LibTorch from source. We currently don't provide any helpers for building it in this repo. 

One thing that we should probably make easier is to provide an env var so that users installing torch from source could specify that they want to build Lib Lantern from source instead of downloading the pre-built binaries. Currently the only way is to clone the repository, build lantern and then install the R package - which is not very simple.

hsbadr:
> One thing that we should probably make easier is to provide an env var so that users installing torch from source could specify that they want to build Lib Lantern from source instead of downloading the pre-built binaries. Currently the only way is to clone the repository, build lantern and then install the R package - which is not very simple.

Yeah, that's exactly what I do. I run a script to clone the sources, build `libtorch` followed by `liblantern` and then copy them to `$TORCH_HOME`. It'd be nice to build `liblantern` from source code when installing `torch`, but right now building them externally keeps the directory structure as expected (all libs and headers are relative to `$TORCH_HOME`). If you build it internally, we need to think about where to install it: should it be copied to `$TORCH_HOME` or in the package directory?

PR #543, opened 2021-05-03
Dev 542

### LINKED ISSUE #542: in nnf_multi_head_attention_forward, return attention weights for each head
Currently, the function averages the attention weights over all the heads before returning:
https://github.com/mlverse/torch/blob/d5cecbf342c857a0fbe9a5a29a09daf1441a706f/R/nnf-activation.R#L733

I'm not sure when it would be useful to have the average, but that's clearly the intended behavior. This function would be more useful to me if it returned the un-averaged weights, i.e.:

`  return(list(attn_output, attn_output_weights)) `
rather than
`  return(list(attn_output, attn_output_weights$sum(dim = 2)/num_heads)) `

Is there any reason *not* to make this change?

I see there's a discussion in pytorch about this, also:
https://github.com/pytorch/pytorch/issues/47583

I suppose there's a desire to keep torch as parallel to pytorch as possible. What's the policy about making changes in torch before they're made in pytorch?

jonathanbratt:
closes #542 

Default behavior is unchanged. 
Added tests + documentation for new parameter.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # torch (development version)
 
 - Fixed bug in `nn_multihead_attention` when q,k,v inputs not all the same. (@jonathanbratt #540)
+- Added parameter to multihead attention module to allow output of unaveraged attention weights. (@jonathanbratt #542)
 
 # torch 0.3.0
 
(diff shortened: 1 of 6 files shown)


dfalbel (approved):

dfalbel:
Looks great @jonathanbratt ! Thanks !

PR #619, opened 2021-07-23
Add CUDA code for sort_vertices.

### LINKED ISSUE #608: Making custom CUDA function for R torch
Hi,

I am translating a pytorch code into R torch but it uses Pytorch functions with CUDA extensions. See cuda_op file in:

[https://github.com/lilanxiao/Rotated_IoU]()

In summary, the code is a loss function (Differentiable) with varying IoU (DIoU, GIoU etc) for rotating (and axis aligned) 3D Boxes. 

I am not sure how to use R torch code with CUDA extensions? 
What do I change in the cuda op file, which consists of CUDA and (C++ or C) wrappers ? 
I am not able to find an examples that demonstrate implementation of a custom CUDA function with R torch (see [ https://github.com/chrischoy/pytorch-custom-cuda-tutorial](url) and [https://discuss.pytorch.org/t/are-there-any-tutorial-for-customizing-functional-using-c-with-cuda/7522](url) for pytorch examples). If one doesn't exist, can anyway demonstrate such an example using the Rotated_IoU code?

Thank you.

dfalbel:
This PR adds changes to lantern so we can add custom CUDA code.
We added the code from https://github.com/lilanxiao/Rotated_IoU use to efficiently sort vertices.

This will eventually fix #608 

TODO:

- [x] Create the Rcpp wrapper around the lantern call.
- [x] Deal with the case when lantern is not built with cuda support. Ie, we need to provide an implementation of `sort_vertices` that works in that case.
- [x] Add comments attributing to the original repository + license.


dominikj2:
Hi @dfalbel, 

Thank you for PR that adds changes to lantern so we can add custom CUDA code. 

Reinstalled torch and found `torch::contrib_sort_vertices()`  . Thank you!

dfalbel:
@dominikj2 Glad it worked! Let me know if you have any problems. It would be nice if at some point we provided a cpu implementation for `contrib_sort_vertices()`. Currently it will only work if all inputs are on the GPU.

PR #644, opened 2021-08-06
Improve cuda detection

### LINKED ISSUE #638: Correctly autodetecting CUDA 11.1 installation
I think this is because `/usr/local/cuda` no longer contains a `version.txt` file.

dfalbel:
Improve cuda detection heuristics by querying nvcc from default locations (even if its not on the path).
Fix #638


jonthegeek (commented):
This looks great! Thanks!

PR #678, opened 2021-08-20
Implement [ for nn_sequential modules.

### LINKED ISSUE #675: Implement `[` for `nn_sequential` modules.
So we can get a subset of the model layers.

mohamed-180:
Fix #675


dfalbel:
Thanks very much @mohamed-180 !

PR #696, opened 2021-09-30
Bugfix/indexing with ellipsis

### LINKED ISSUE #695: Indexing error when using `..`
I think there is another issue now, 

```
a <- torch_randn(c(3,4,2))
c <- torch_randn(c(3,4,2))
b <- c(1,2,4)
a[..,torch_tensor(as.integer(b)),]
c[..,torch_tensor(as.integer(b)),]
b
```

`Error in Tensor_slice(tensor$ptr, environment(), drop = drop, mask = .d): index 3 is out of bounds for dimension 0 with size 2
Exception raised from get at aten/src/ATen/native/cpu/IndexKernel.cpp.AVX2.cpp:42 (most recent call first)`

The error does not occur if we do not use the `..` annotation.

_Originally posted by @emauryg in https://github.com/mlverse/torch/issues/691#issuecomment-927501469_

dfalbel:
Fix #695


dfalbel:
The CI error happens because of a regression in knitr that has already been solved in the 1.36 release.

PR #697, opened 2021-09-30
Feature/custom class

### LINKED ISSUE #682: Datasets constructors should have an specific class
For example:

``` r
x <- torch::dataset(
    initialize = function() {},
    .getitem = function(i) {
      torch_randn(10, 10)
    },
    .length = function() {10}
  )
  
class(x)
#> [1] "function"
```

dfalbel:
Fix #682 

Datasets constructors now have a custom class so one can easily check if an object is a dataset constructor. 
We also added an improved print method.


skeydan (approved):
That's definitely more useful :-)

PR #698, opened 2021-09-30
Improve the optimizer documentation.

### LINKED ISSUE #683: Add note about optimizer creation
We should a note in the docs like:

> If you need to move a model to GPU via .cuda(), please do so before constructing optimizers for it. Parameters of a model after .cuda() will be different objects with those before the call.
In general, you should make sure that optimized parameters live in consistent locations when optimizers are constructed and used.

https://pytorch.org/docs/master/optim.html 
cc @jonathanbratt

dfalbel:
Include a warning related to the order of the creation of the optimizer in torch.
Fix #683


skeydan:
@dfalbel somehow I can't add a review  - nothing happens when I click on "add your review" ...

I'll just comment here:

> If you need to move a model to GPU via `$cuda()`, please do so before constructing optimizers for it. Parameters of a model after `$cuda()` will be different objects with those before the call.

--> change _with those_ -> _from those_

> In general, you should make sure that optimized parameters live in consistent locations when optimizers are constructed and used.

--> this sentence is not very clear, it seems to me. I'd either remove it, or rephrase in a more concrete way.

dfalbel:
Thanks @skeydan 
Rewrote the second sentence. WHat do you think?

skeydan:
Aha ok, now I get it! How about this, then:

In general, you should make sure that the objects pointed to by model parameters subject to optimization remain the same over the whole lifecycle of optimizer creation and usage.

PR #699, opened 2021-10-01
Make `nn_sequential` a bare `nn_module`

### LINKED ISSUE #674: `nn_sequential` should be a bare `nn_module` not a function
So we can inherit from it, example: https://github.com/pytorch/vision/blob/master/torchvision/models/mobilenetv2.py#L32

dfalbel:
This PR makes `nn_sequential` a bare `nn_module`. This allow inheriting from `nn_sequential` that might be useful in some scenarios. 

We no longer support the `name` argument, but its easy to modify the name via inheritance. See the modified test case.
Fix #674

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 - Fixed bug when indexing with numeric vectors. (#693, @mohamed-180)
 - Fixed bug when indexing tensors with ellipsis and a tensor. (#696)
 - Improved optimizer documentation by adding a 'Warning' regarding the creation and usage order. (#698)
-- `nn_sequential` is now a bare `nn_module`, allowing to easily inherit from it. This is a breaking change if you used the `name` argument. The `name` behavior can be achieved by inheriting see the tests in the PR. (#699)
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -7,6 +7,7 @@
 - Fixed bug when indexing with numeric vectors. (#693, @mohamed-180)
 - Fixed bug when indexing tensors with ellipsis and a tensor. (#696)
 - Improved optimizer documentation by adding a 'Warning' regarding the creation and usage order. (#698)
+- `nn_sequential` is now a bare `nn_module`, allowing to easily inherit from it. This is a breaking change if you used the `name` argument. The `name` behavior can be achieved by inheriting see the tests in the PR. (#699)
```
skeydan, line 10:
```suggestion
- `nn_sequential` is now a bare `nn_module`, allowing to easily inherit from it. This is a breaking change if you used the `name` argument. The `name` behavior can be achieved by subclassing; see the tests in the PR. (#699)
```

skeydan (approved):
LGTM!

PR #701, opened 2021-10-01
Persist classes from parents in subclasses

### LINKED ISSUE #700: `nn_module` inheritance should keep `inherit` class
For example:

```
nn2 <- nn_module(classname = "hello", inherit = nn_linear)
class(nn2(10, 10))
```

Should return `hello, nn_linear, nn_module`, not `hello, nn_module`.

dfalbel:
Allows persisting the inherited classes. It's expected that subclasses also inherits the classes from their parents.
Fix #700 

This is specially useful when inheriting from `nn_sequential`, allowed by #699 . Since `nn_sequential` implements other methods for the module, eg `[`, and `length`.

Merge after #699


skeydan (approved):
LGTM!

Issue #708, opened 2021-10-08
torch_sort produces index 0-based but should be 1-based for R 


dominikj2:
Hi,

Does this need to be corrected to make torch_sort results 1-based for R?
```
torch_sort(torch_tensor(c(3,4,2,1)))[[2]]
torch_tensor
 3
 2
 0
 1
[ CPULongType{4} ]
```

Thanks

dfalbel:
Fixed in https://github.com/mlverse/torch/pull/709

PR #709, opened 2021-10-08
torch_sort produces index 0-based but should be 1-based for R #708

mohamed-180:
it becomes 
```
torch_sort(torch_tensor(c(3,4,2,1)))[[2]]
torch_tensor
 4
 3
 1
2
[ CPULongType{4} ]
```


dfalbel:
THanks @mohamed-180 ! I'm waiting for a patch release of torch to get accepted on CRAN before merging this PR.

R/with-indices.R
```diff
@@ -88,4 +88,10 @@ torch_cross_entropy_loss <- function(
   target <- target$sub(1L, 1L)
   .torch_cross_entropy_loss(self = self, target = target, weight = weight,
                             reduction = reduction, ignore_index = ignore_index)
-}
\ No newline at end of file
+}
+
+torch_sort <- function(self, dim = -1L, descending = FALSE, stable = TRUE) {
+  out <- .torch_sort(self, dim, descending = FALSE, stable )
```
hsbadr, line 94: This doesn't respect the `descending` argument passed to `torch_sort()`.

mohamed-180: That is true , thanks @hsbadr .

PR #713, opened 2021-10-12
implement jit_save_for_mobile

skeydan:
The tests do not check whether a `LiteModuleLoader` can actually load the modules; instead, they just test if formally, the code works.
I can also confirm that the saved archives contain a `bytecode.pkl` file.

One thing that probably wasn't needed is the method for `ScriptFunction`, as these probably won't be exported standalone - but I guess it doesn't hurt to have it, either ...


skeydan:
Hm, I committed an update of NEWS.md, but you probably meant an update of a source file in lantern?

PR #718, opened 2021-10-20
Feature/install headers

dfalbel:
- [x] Install libtorch headers when installing libtorch.
- [x] Install the R package headers along with the package so other packages can use them.
- [x] ~~`library.dynam` instead of the `useDynLib` directive. This allows us to set `local=FALSE` which is necessary for other packages using torch symbols to be able to find them.~~
- [x] Add a C api that is now exported via the R `R_RegisterCCallable` and `R_GetCCallable` mechanism.
- [x] Refactor torch types into namespaces and with more intuitive names.


R/install.R
```diff
@@ -33,16 +33,16 @@ install_config <- list(
     "10.2" = list(
       "linux" = list(
         "libtorch" = list(
+          path = "libtorch/",
```
hsbadr, line 36: Since `path` is constant, it's cleaner to remove it from all lists. This requires updating `source_path` in `lantern_install_libs()` and `lantern_install_lib()`.

R/install.R
```diff
@@ -378,3 +401,4 @@ get_install_libs_url <- function(version = "1.9.1", type = install_type(version
   list(libtorch = libtorch, liblantern = liblantern)
 }
 
+
```
hsbadr, line 404: I'd like to remove all whitespaces and empty lines, after merging the pending PRs.

dfalbel: That would be great!

R/install.R
dfalbel: True! I'll fix this

dfalbel: Turns out this is implicitly an empty string for lantern configurations, so I had to revert that change.

R/install.R
hsbadr: I'm on it. It will be a relatively big change for code cleanup, styling, and formatting.

PR #730, opened 2021-11-04
exported is_torch_tensor. also fixed typo in autograd_function docs

### LINKED ISSUE #728: export `is_torch_tensor()`?
I always end up writing an `is_torch_tensor()`  as a utility function for most of my `torch` projects, because I get tired of typing `inherits(x, "torch_tensor")` (which cannot autocomplete). Would you consider exporting this function? I find it pretty handy.

rdinnager:
Mostly just exports `is_torch_tensor()`. I also happened to notice a typo in the `autograd_function()` example in the documentation, which I fixed (I can confirm the typo caused my r session to crash when I ran it and then tried to get the gradient 😅 , but it works after fixing it 👍 ). Fixes #728


dfalbel:
THanks @rdinnager ! I am currently waiting for a patch CRAN release before merging changes, but will merge if CRAN requests additional changes from that patch release. Sorry for the delay.

PR #742, opened 2021-12-03
add .getbatch to dataset subsets

egillax:
I added a .getbatch method to the dataset_subset so it would work with batched indices.


dfalbel (commented):
Could we add a unit test for the case where the dataset being subset doesn't have a `.getbatch` method? I think we might need to check if it exists and otherwise delegate to `.getitem()`.

egillax:
Hi @dfalbel,

I added a test for the case where the dataset being subset only has a .getitem method. It passes for me locally. Let me know if this is what you had in mind.

dfalbel:
Thanks @egillax ! I did a small change to make it work when `.getitem` fail when called with more than 1 item.

PR #743, opened 2021-12-03
implemented call_torch_function(). fixes #731

### LINKED ISSUE #731: An idea: a function to retrieve unexported `torch_` prefixed functions? (because I need one of them)
I just want to start this (long) issue by saying what an incredible package this is and what a resource to the R / machine learning / statistical inference community! I have been studying the package design over the last few months and have frequently been amazed at what a fascinating build system it is.

In the process of studying this package and also trying to port some complicated Pytorch code I have noted that there is a large number of autogenerated R functions that interface with underlying `libtorch` c++ functions. These are all prefixed with `torch_` and many are not exported (by design presumably, as they are mostly meant to be used in other exported functions). However, currently many of these functions are actually totally unused anywhere in any exported function. This made me wonder if there would be value in a function designed specifically for advanced users that allowed access to these functions, without having to export them all and clutter the public namespace? So something along the lines of an `call_torch_function(func_name, args)`, which would construct a call to the unexported function and return the result. Warnings would have to be made to users that you would really have to know what you are doing to use this (and it is at your own risk), since all the functions have not been wrapped in an R-friendly way (e.g. changing indexing to 1-based instead of 0-based, etc.), and many also are presumably untested (?).

I have a particular use-case in mind. In the Pytorch project I am porting they use a call directly to an operation `cudnn_convolution_backward_weight`. After digging around for awhile I discovered `torch:::torch_cudnn_convolution_backward_weight()` exists as an unexported function in `torch` (with the same function signature). Now, I realise I could request this function be exported, but then it would kind of hang there by itself in the package, as something hardly anyone would use. And then the question is, should also `torch::torch_cudnn_convolution_backward_input()` be exported for consistency? What about all the other `cudnn_convolution_` operations? These functions are potentially useful, but only in limited circumstances, and usually only by very advanced users. That is why I think a function that could send a call to any of these might make sense. Note that I am not thinking it should allow access to any non-exported function in the package, just the `torch_` prefixed ones, since these are all in `Declarations.yaml`, which could be the criteria for allowing access to advanced users (or `torch` developers if you prefer). A function like this would also allow R package developers to write their own user friendly wrappers for `libtorch` functions that are currently unwrapped in `torch` itself.

Anyway, just a (possibly crazy) thought. Feel free to shoot it down immediately. In that case though, can I get that `torch_cudnn_convolution_backward_weight()` function exported? Happy to make a PR for it (or for the above idea). Thanks again for the great package.

rdinnager:
Okay, here is `call_torch_function()`. Please let me know what you think. I have used `rlang::list2()` to collect the `...` because I find it useful to use `!!!` with this function, but could easily revert to `base::list()` if you don't want to use `rlang` here.


dfalbel:
Thanks @rdinnager !

PR #754, opened 2022-01-11
[WIP] attempt to simplify concurrency-related constructs in autograd.cpp

yitao-li:
Signed-off-by: Yitao Li <yitao@rstudio.com>


src/autograd.cpp
```diff
@@ -18,6 +19,23 @@
...
+    // If we are in the main thread, then spawn another thread to carry out
+    // the `backward` task asynchronously.
+    // If we are already outside the main thread), then run `backward`
+    // synchronously in the current thread.
+    std::this_thead::get_id() == main_thread_id() ?
+      std::launch::async : std::launch::deferred
+  );
+
+  std::async(policy, std::move(backward)).get();
+}
```
yitao-li, line 35: I'm hazarding a guess that we won't really need any of the task deque data structures (as indicated by the numerous red rectangles in the rest of this PR), and all we need is simply this... but I could be completely wrong :sweat_smile:

src/autograd.cpp
```diff
@@ -26,15 +44,15 @@ void cpp_autograd_set_grad_mode (bool enabled) {
 // [[Rcpp::export]]
 bool cpp_autograd_is_enabled()
 {
-  return lantern_autograd_is_enabled();
```
yitao-li, line 29: Please ignore whitespace changes for now. Sorry.

You can click the gear icon on top, and then tick the "Hide whitespace" check box, and then "apply and reload".

src/autograd.cpp
dfalbel: I think at least one event loop will be needed because the R function that's called at some point by `lantern_autograd_backward` must be somehow scheduled to be executed in the main thread.

yitao-li:
Gonna try a slightly different approach in a separate PR.

PR #755, opened 2022-01-12
 simplify concurrency-related constructs in autograd.cpp

yitao-li:


inst/include/utils.h
```diff
@@ -1,5 +1,75 @@
...
+#include <stack>
+#include <mutex>
+#include <condition_variable>
+#include <future>
+#include <functional>
+#include <utility>
+#include <type_traits>
+
+template <typename F>
+class ScopeGuard {
```
yitao-li, line 12: I expect this to be part of the C++ standard in near future (e.g., see https://en.cppreference.com/w/cpp/experimental/scope_exit).

This is just my straw-man implementation of the scope guard concept.
There is also a much more advanced version implemented by advanced C++ gurus doing advanced C++ full-time in https://github.com/facebook/folly/blob/47eda458e9763093e0bc1baa3bebef6aeb816513/folly/ScopeGuard.h#L1

inst/include/utils.h
```diff
@@ -1,5 +1,75 @@
...
+class EventLoop {
+ public:
+  EventLoop() = default;
+  void run() {
+    while (true) {
+      std::packaged_task<T()> fn;
+      {
+        std::unique_lock<std::mutex> lock(mtx_);
+        if (tasks_.empty()) {
+          cv_.wait(lock, [this] { return !tasks_.empty(); });
```
yitao-li, line 41: - Minor improvement: using condition variable to make this thread up after the task queue goes from empty to non-empt.y
- Another related minor improvement: not calling `unlock()` or `lock()` manually here (as doing so here while also ensuring exception-safety is a bit error-prone)

src/autograd.cpp
```diff
@@ -57,145 +57,82 @@ bool cpp_tensor_requires_grad (torch::Tensor self) {
...
-        task();
-        lock.lock();
-      }
-    }
-    
+EventLoop<void*> gTasks;
+EventLoop<void> gBackwardTasks;
+
+void schedule_backward_task(std::packaged_task<void()>&& task) {
+  if (std::this_thread::get_id() == main_thread_id()) {
```
yitao-li, line 66: Another simplification: using this instead of `n_tasks` to decide when to launch something in a new thread and when to continue in the current thread.

inst/include/utils.h
```diff
@@ -1,5 +1,75 @@
...
+  explicit ScopeGuard(F&& f) noexcept : f_(std::forward<F>(f)) {}
+  ~ScopeGuard() noexcept { f_(); }
+ private:
+  typename std::decay<F>::type f_;
+};
+
+// TODO:
+// create a factory function for ScopeGuard that disallows the returned
+// ScopeGuard object to be discarded immediately (but we need to do so in a
+// portable manner, by using the [[nodiscard]] attribute if possible, or
```
yitao-li, line 23: Having `[[nodiscard]]` here is actually quite important because the ScopeGuard object is supposed to be destroyed when out of scope rather than immediately, because if someone were to mistakenly have

```
makeScopeGuard(fn);
```

or

```
ScopeGuard(fn);
```

instead, then the `ScopeGuard` object gets destroyed after the `;`, which would make it not work as expected.

The `[[nodiscard]]` attribute is only standardized in C++-17 though. I'm not sure whether there is a portable solution here if we want to stick with C++-11.

yitao-li:
There is a mysterious build RMD failure on Windows after this change, but aside from that looks like it's working fine on Linux+CPU and Linux+GPU platforms.

inst/include/utils.h
```diff
@@ -1,5 +1,75 @@
...
+      std::packaged_task<T()> fn;
+      {
+        std::unique_lock<std::mutex> lock(mtx_);
+        if (tasks_.empty()) {
+          cv_.wait(lock, [this] { return !tasks_.empty(); });
+        }
+        fn = std::move(tasks_.front());
+        tasks_.pop_front();
+      }
+      if (!fn.valid()) {
```
dfalbel, line 46: What's the meaning of `fn.valid()` here? Does it mean that the task is done?

inst/include/utils.h
```diff
@@ -1,5 +1,75 @@
...
+    {
+      std::unique_lock<std::mutex> lock(mtx_);
+      tasks_.emplace_front(std::move(task));
+    }
+    cv_.notify_one();
+  }
+  void stopWhenEmpty() {
+    {
+      std::unique_lock<std::mutex> lock(mtx_);
+      tasks_.emplace_back();
```
dfalbel, line 63: Ohh I see, so this puts an empty task, that will later become `fn`, and if `!fn.valid()` we stop the process.

src/autograd.cpp
```diff
@@ -57,145 +57,90 @@ bool cpp_tensor_requires_grad (torch::Tensor self) {
...
-    catch (...)
-    {
-      event_loop_running = false;
-      n_tasks = n_tasks - 1;
-      throw;
-    }
-    
-    event_loop_running = false;
-    n_tasks = n_tasks - 1;
+    auto sg = makeScopeGuard([] { gTasks.stopWhenEmpty(); });
```
dfalbel, line 95: Just to confirm: With the new implementation the event loop won't stop until this `sg` object is deleted. And it's always eventually deleted at the end of the execution of this `backward` lambda function. 

If that's correct. I have a question: why not simply ending the event loop when the task list is empty? why do we prefer waiting for this object to be deleted?

src/autograd.cpp
```diff
@@ -215,35 +159,23 @@ unsigned int cpp_tensor_register_hook (Rcpp::XPtr<XPtrTorchTensor> self, Rcpp::F
...
-        }
-        
-      } 
-    } while (status != std::future_status::ready); 
-    
-    return result.get();
+        gBackwardTasks.stopWhenEmpty();
+        gBackwardTasks.run();
+      }
+      std::this_thread::yield();
```
dfalbel, line 173: From docs: Provides a hint to the implementation to reschedule the execution of threads, allowing other threads to run.

What does this mean?

dfalbel (commented):
That's really great @yitao-li ! Learning a lot from this PR!
Left a few questions about things that are not yet completely clear to me.

inst/include/utils.h
yitao-li: Yep. `valid()` will return `false` if and only if the underlying function object in the packaged task is `nullptr` (see the implementation of `stopWhenEmpty()` below)

src/autograd.cpp
yitao-li: That's right.

Re: `stopWhenEmpty()` and why -- I was not sure whether stopping the event loop as soon as it becomes empty is the correct thing to do. In particular, if I look at the existing implementation (before this change), it looks like there could be a case where the event loop becomes momentarily empty, but because `event_loop_running` is still set to `true`, we keep polling the event loop for further tasks until `event_loop_running` is set to `false`. So, this implementation preserves that behavior.

So concretely, that means some synchronous function in this lambda can place some more tasks in the event loop (as long as they do so before `event_loop_running` is set to `false` at the end), and then those tasks are guaranteed to be run following the completion of the lambda itself (in other words, the existing implementation ensures some additional tasks can be deferred to after the completion of this lambda if necessary).

src/autograd.cpp
yitao-li: It means suggesting this thread to momentarily give up its place on the CPU and possibly allow the OS to schedule other threads waiting for CPU (e.g., calling `usleep(1)` would have the same effect, but is most likely less efficient, whereas `std::this_thread::yield()` is possibly more efficient, and also expresses the intention better).

Why do it: because without `yield()` here, the future status would be polled in a tight loop, which means other threads do not get a chance to be scheduled on the same core while this loop is running. So, other threads could be potentially starving if there is no free core aside from the current one executing this loop.

dfalbel (approved):
Look great!

PR #760, opened 2022-01-18
Disable warnings in `dir.create()` and remove `fs` dependency

hsbadr:
This PR disables warnings in `dir.create()` using `showWarnings = FALSE` argument to avoid using `suppressWarnings()` nor checking if the directory exists. It also simplifies `lantern` build and removes `fs` dependency by replacing `fs::dir_create()` with `dir.create()`. The `recursive = TRUE` argument makes `dir.create()` similar to the Unix command `mkdir -p`.


dfalbel:
Thanks @hsbadr !

PR #780, opened 2022-02-10
Expanded the loading data vignette

jnolis:
I expanded the load data vignette to give more of a gentle introduction for new users. I also added a part at the end about optimizing data loads that mentions getbatch and avoiding datasets entirely.

I also added a sentence or two to the `dataset` docs, although I wasn't sure how to have a hyperlink to the vignette (I'm not an R package expert)


R/utils-data.R
```diff
@@ -39,12 +39,13 @@ get_init <- function(x) {
...
 #' By default datasets are iterated by returning each observation/item individually.
-#' Sometimes it's possible to have an optimized implementation to take a batch
+#' Often it's possible to have an optimized implementation to take a batch
 #' of observations (eg, subsetting a tensor by multiple indexes at once is faster than
 #' subsetting once for each index), in this case you can implement a `.getbatch` method
 #' that will be used instead of `.getitem` when getting a batch of observations within
-#' the dataloader. `.getbatch` must work for batches of size larger or equal to 1.
-#'
+#' the dataloader. `.getbatch` must work for batches of size larger or equal to 1. For more
+#' on this see the the vignette on data loading.
```
jnolis, line 47: This ideally would have a link to the vignette

jnolis:
Also if this is helpful I bet I could do it for other pages too!

R/utils-data.R
dfalbel, line 47: This will make at least the pkgdown website to create a link:

```suggestion
#' on this see the the `vignette("loading-data")`.
```

Up to you if you prefer adding this or not. I think most users will read it in the Rd format.

Issue #797, opened 2022-03-03
Detect CUDA version without version.txt?


SvenVw:
I am trying to install the gpu version of torch on my windows machine. However, torch does not to recognize that CUDA 11.1 is installed on my machine.
When looking at the source code I see that torch checks for a `version.txt` file located at `cuda_path`:
https://github.com/mlverse/torch/blob/0afe34f8fbf13a7ff2a13c342544bf9a5e0b8c84/R/install.R#L243
However, it seems that from CUDA 11 onwards this file is not supplied anymore. Could another method be used to detect the CUDA version? For example: https://github.com/llvm/llvm-project/commit/e7fe125b776bf08d95e60ff3354a5c836218a0e6

Thanks in advance!

dfalbel:
Are you on Linux or Windows?

Yeah, we should definitely implement something like this. Looks like parsing `cuda.h` is a reasonable approach. I think we already handle on Linux by querying nvcc:

https://github.com/mlverse/torch/blob/0afe34f8fbf13a7ff2a13c342544bf9a5e0b8c84/R/install.R#L302-L304

I think both approaches should be fine. Would you like to contribute a PR?

In the meantime, you can manually set the cuda version with `Sys.setenv(CUDA="11.1")` too to avoid the automatic detection.

SvenVw:
Yes I am on Windows. Parsing `cuda.h` seems a little bit difficult to me, but querying nvcc should be possible. I will try to contribute a PR for this issue

dfalbel:
Awesome! That's very helpful. I don't have a Windows machine with a GPU to experiment, so it's hard to make sure it works.

dfalbel:
Closed in #798

PR #798, opened 2022-03-04
On Windows query cuda_version from nvcc instead from version.txt

SvenVw:
Hi, this PR solves the issue #797 on windows that the `version.txt` file is not present anymore from CUDA 11 onwards. It now uses the same function as already is used on Linux to query the `cuda_version` from the nvcc command instead of parsing `version.txt`. However, instead of the using environment variable `CUDA_HOME`  (which is used on linux) `CUDA_PATH` is used and it refers to the nvcc executable.

Testing it on my windows machine with CUDA 11.1 resulted in a successful installation of `torch` for gpu.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -239,11 +239,10 @@ install_type_windows <- function(version) {
   cuda_version <- NULL
   cuda_path <- Sys.getenv("CUDA_PATH")
 
-  if (nzchar(cuda_path)) {
-    versions_file <- file.path(cuda_path, "version.txt")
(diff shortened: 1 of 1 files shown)


dfalbel:
Thanks very much @SvenVw ! I made a small modification because I think we should keep looking for the version file and if not found query NVCC. Could you check that this still works?

SvenVw:
Ahh, yes that is best of both worlds ;). I can confirm that it still works

dfalbel:
Cool, thanks very much!

PR #803, opened 2022-03-30
Dataloaders with socket conection

dfalbel:
Passing data between the workers can now be accelerated by using socket connections.
```
> bench_ds <- dataset(
+   initialize = function(size_of_tensor) {
+     self$size_of_tensor <- size_of_tensor
+   },
+   .getitem = function(i) {
+     list(x = torch_randn(self$size_of_tensor))
+   },
+   .length = function() {32*20}
+ )
> system.time({
+   withr::with_options(list(torch.dataloader_use_socket_con = TRUE), {
+     dl <- dataloader(bench_ds(256*256*3), batch_size = 32, num_workers = 4)
+     coro::loop(for(x in dl) {
+       k <- x
+     })
+   })
+ })
   user  system elapsed 
 48.416   2.419  58.796 
> system.time({
+   withr::with_options(list(torch.dataloader_use_socket_con = FALSE), {
+     dl <- dataloader(bench_ds(256*256*3), batch_size = 32, num_workers = 4)
+     coro::loop(for(x in dl) {
+       k <- x
+     })
+   })
+ })
   user  system elapsed 
 61.417   2.273  74.165 
```

Next steps should be investing in `torch_save` speedups, specially, if we could avoid the compression.
Perhaps, also finding a different way of serializing the tensor.

PR #806, opened 2022-03-31
Fix headers after #803

hsbadr:
This also fixes building [`lltm`](https://github.com/mlverse/lltm) after #803.


dfalbel:
Thank you @hsbadr !

dfalbel:
I should to find a way to test this on CI!

PR #827, opened 2022-05-12
Embedding fix

egillax:
Hi @dfalbel , 

I added a nn module for embedding bags since I need that for my models. As well added support for 2d input to nnf_embedding_bag since I needed those as well. I looked at the python code and the current implementation of the nn_embedding for guidance. I also wrote some basic tests, but maybe they could be more extensive.


dfalbel:
Thanks very much @egillax !

PR #828, opened 2022-05-13
Add batch_first arg for attention.

jonthegeek:
I haven't tested this; Rstudio crashed when I tried to build torch locally. Theoretically it's straightforward, but I'd definitely like to test it before I call this done. Any help is appreciated!

### DIFF
--- a/R/nn-activation.R
+++ b/R/nn-activation.R
@@ -1,7 +1,7 @@
 #' @include nn.R
 NULL
 
-#' Threshoold module
+#' Threshold module
 #'
 #' Thresholds each element of the input Tensor.
 #'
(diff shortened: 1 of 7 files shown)


dfalbel:
To build torch locally you will need to run `source('tools/buildlantern.R')`. THen `devtools::load_all()` should just work.

jonthegeek:
I'm getting a bunch of errors like this: `c++.exe: error: C:/rtools42/permissive: No such file or directory`

Evidently some make vars are getting misinterpreted.

It may be time for me to finally learn about this WSL thing to run Linux on a Windows machine...

dfalbel:
Hmm, I haven't seen those errors. Might be a problem with Rtools? 
Or maybe you will need to install the visual studio c++ compiler . I'll review this PR tomorrow!

jonthegeek:
I was able to install this branch with install_github, so it's specifically something about sourcing that makes rtools act up. I'll keep poking at it, but I can use this workaround to try to actually sort out tests tomorrow.

dfalbel:
Fix #829

PR #836, opened 2022-05-26
add lr_reduce_on_plateau

egillax:
Added a new learning rate scheduler, lr_reduce_on_plateau with tests and documentation.

Modelled it after the pytorch version. Also made tests and made sure the test cases matched with the output from pytorch.


dfalbel (approved):
Looks good to me! Thank you very much @egillax !

Issue #736, opened 2021-11-14
Is it possible to detect if tensor is sparse or dense?


egillax:
I'm trying to make a custom collate_fn where some of my data is sparse and some dense. With the sparse data I want to stack and then use to_dense() but with the dense I only want to stack.

In pytorch there is the function [is_sparse](https://github.com/pytorch/pytorch/blob/1022443168b5fad55bbd03d087abf574c9d2e9df/torch/_linalg_utils.py#L11) which relies on the tensor.layout property. 

It doesn't seem like the layout property is exposed here. Is it possible to expose it so I can detect if the tensor is sparse? Or is there already some way to do it?

Regards,
Egill

dfalbel:
@egillax sorry, this is currently not possible, but we will make sure to implement it for the version.

rdinnager:
It turns out this would be handy to have for autograd functional (#738 ), so I can add this functionality as part of the PR for that. It should be a pretty straightforward update to Lantern, since the Torch C++ API provides `is_sparse()` (so we just need a basic wrapper). Or would it be better to make a separate PR for this?

dfalbel:
Cool! No need to create a separate PR.

rdinnager:
Just adding a reference to the `is_sparse()` function in C++ so I don't forget where it is (curious that the Python interface doesn't use this and instead uses the `layout` property, but since R doesn't yet implement `layout`, going this route seems easier for now).
https://github.com/pytorch/pytorch/blob/0032fa772588cd20f2b85324bdeb31ac497bbf29/c10/core/TensorImpl.h#L813-L820

PR #871, opened 2022-08-21
add is_sparse method for tensors

egillax:
solves #736 

Full credit goes to @rdinnager since I got this from his branch [here](https://github.com/rdinnager/torch/tree/feature/autograd_functional). I just needed this functionality again and there's been no movement on that branch since january.

This passes all tests locally


rdinnager:
Hi @egillax, glad this will get some use. I got really busy with other stuff and don't know when I will be able to get back to implementing functional (if ever), so feel free to use anything from that branch (not that I got very far with it).

dfalbel:
Sorry @egillax ! I ended up re-implementing it here: https://github.com/mlverse/torch/pull/903 to avoid conflicts when merging.

PR #880, opened 2022-09-14
Fix lenght typos.

jonthegeek:
I saw one of these while reading a help doc so I figured I might as well find-and-replace them all.


dfalbel:
THanks very much @jonthegeek !

PR #899, opened 2022-10-09
Minor updates after #893 (LibTorch 1.12.1)

hsbadr:
Rename the declarations file and update the default version for `torchgen`, following #893.


dfalbel:
Thanks!

PR #904, opened 2022-10-14
Subset classes

### LINKED ISSUE #881: dataset_subset hides dataset subclass
I expected `dataset_subset` to keep dataset properties, but it hides them. Could we either sort out a way to elevate the dataset to the "top" of that object, or at least give the new dataset a subclass so methods can know to look down a level for "real" dataset properties?

``` r
library(torch)
minimal_dataset <- dataset(
  "minimal",
  initialize = function() {
    self$data <- torch_tensor(1:5)
  },
  .length = function() {
    self$data$size()[[1]]
  }
)
testing <- minimal_dataset()
class(testing)
#> [1] "minimal" "dataset" "R6"
testing_sub <- dataset_subset(testing, 1:2)
class(testing_sub)
#> [1] "dataset" "R6"
```

<sup>Created on 2022-09-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dfalbel:
Fix #881 

@jonthegeek We now append `_subset` to the base dataset class, so you can easily write S3 methods for it if necessary.
I don't think there's a safe way to forward the base object methods to the subset as those methods won't know about the selected index and etc. But at least you can now write a custom `$` method for your dataset.

### DIFF
--- a/R/utils-data.R
+++ b/R/utils-data.R
@@ -148,6 +148,9 @@ dataset_subset <- dataset(
     if (!is.null(dataset$.getbatch)) {
       self$.getbatch <- self$.getitem
     }
+    classes <- class(dataset)
+    classes_to_append <- classes[classes != "R6"]
+    class(self) <- c(paste0(classes_to_append, "_subset"), classes)
   },
   .getitem = function(idx) {
     return(self$dataset[self$indices[idx]])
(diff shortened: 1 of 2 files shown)


jonthegeek:
Perfect! This is effectively what I'm currently doing, but having it built in will make the code much cleaner (and safer)!

PR #936, opened 2022-12-06
Added a verbose mode on Windows installation to help find CUDA

cyrilmory:
Added a verbose mode on Windows installation
Switched to reading "version.json" instead of "version.txt" in CUDA_PATH to find CUDA version on Windows

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -235,20 +235,30 @@ lantern_install_libs <- function(version, type, install_path, install_config) {
 }
 
 install_type_windows <- function(version) {
+  verbose <- Sys.getenv("TORCH_VERBOSE_INSTALL")
   cuda_version <- NULL
   cuda_path <- Sys.getenv("CUDA_PATH")
 
   if (nzchar(cuda_path)) {
-    versions_file <- file.path(cuda_path, "version.txt")
+    if (verbose == "1") { message(paste0("CUDA_PATH environment variable is set to ", cuda_path)) }
+    versions_file <- file.path(cuda_path, "version.json")
     if (file.exists(versions_file)) {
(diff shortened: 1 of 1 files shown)


R/install.R
```diff
@@ -235,20 +235,30 @@ lantern_install_libs <- function(version, type, install_path, install_config) {
...
 
 install_type_windows <- function(version) {
+  verbose <- Sys.getenv("TORCH_VERBOSE_INSTALL")
   cuda_version <- NULL
   cuda_path <- Sys.getenv("CUDA_PATH")
 
   if (nzchar(cuda_path)) {
-    versions_file <- file.path(cuda_path, "version.txt")
+    if (verbose == "1") { message(paste0("CUDA_PATH environment variable is set to ", cuda_path)) }
+    versions_file <- file.path(cuda_path, "version.json")
```
dfalbel, line 244: I think we need test for both files (`version.txt` and  `version.json`) to make sure older versions are also supported.

R/install.R
```diff
@@ -235,20 +235,30 @@ lantern_install_libs <- function(version, type, install_path, install_config) {
...
     if (file.exists(versions_file)) {
-      cuda_version <- gsub("CUDA Version |\\.[0-9]+$", "", readLines(versions_file))
+      if (verbose == "1") { message(paste0(cuda_path, "\\version.json found, extracting CUDA SDK version from it.")) }
+      cuda_version <- stringr::str_extract(readLines(versions_file, n=4)[4], "[0-9]+.[0-9]+")
```
dfalbel, line 247: I don't think we want to take a `stringr` dependency here. Can we rewrite with base R?

dfalbel (commented):
Thanks very much @cyrilmory ! Looks great! I made some inline comments, let me know what you think.

PR #943, opened 2022-12-13
docs: fix nn_glu class name and document

sebffischer:


dfalbel:
Hi @sebffischer! Thanks for the PR! THe PR ended up with too many conflicts, do you mind just commiting the change in `R/nn-activation.R` and we will take are of documenting after the merge is done?

Issue #935, opened 2022-12-05
single-argument torch.where returns 0-based indices


skeydan:
```
> t <- torch_arange(1,5)
> torch_where(t < 2)
[[1]]
torch_tensor
 0
[ CPULongType{1} ]
```

PR #944, opened 2022-12-13
export fftfreq

skeydan:
Hi Daniel,

can you please take a look at this?
It successfully exports the function, but I was not sure how to handle the keyword arguments (I think they're subsumed under `TensorOptions`, but still, documented as usual?).

Also, `document.R` [still ]behaves weirdly for me; always crashes on first execution. Do you have that too?

Finally, I've gotten `.Rd` files (which have not been added [yet]), which do three different kinds of things:

1) add back the "if torch installed" line. I guess this one is removed when not using `document.R`?

2) remove this (in `linalg_cond.R`):

```
-\tabular{ll}{
-   \code{p} \tab matrix norm \cr
-   \code{NULL} \tab \code{2}-norm (largest singular value) \cr
-   \code{'fro'} \tab Frobenius norm \cr
-   \code{'nuc'} \tab nuclear norm \cr
-   \code{Inf} \tab \code{max(sum(abs(x), dim=2))} \cr
-   \code{-Inf} \tab \code{min(sum(abs(x), dim=2))} \cr
-   \code{1} \tab \code{max(sum(abs(x), dim=1))} \cr
-   \code{-1} \tab \code{min(sum(abs(x), dim=1))} \cr
-   \code{2} \tab largest singular value \cr
-   \code{-2} \tab smallest singular value \cr
-}
```

3) remove this (in various optimizer files)

```
-\section{Warning}{
-If you need to move a model to GPU via \verb{$cuda()}, please do so before
-constructing optimizers for it. Parameters of a model after \verb{$cuda()}
-will be different objects from those before the call. In general, you
-should make sure that the objects pointed to by model parameters subject
-to optimization remain the same over the whole lifecycle of optimizer
-creation and usage.
-}
```

Should I add these files? 

Also, I see we point to `document.R` in the contributing guide. However, perhaps visibility would be better if this was made part of the previous section (workflow)?

Thanks!


dfalbel (commented):
Hey @skeydan !

Thanks for the PR!!

- It would be great to add a simple test for that function. Again, I know that's not the case for the entire codebase, but should be our goal when adding new functionality.

- I don't know what causes the difference in the `document.R` run. I think it's better to not add those changes to this PR. Better to use a different PR for this.

- `document.R` just register some S3 methods then run `devtools::document()`. In general the issues are related to not running it in a clean session, it's good practice to restart the session and then run it. At some point this was caused by a bug in `pkgload::load_all()`.

- Yeah, it would be great to improve the contributing docs. Feel free to move sessions, and improve the text, that's always helpful!

R/gen-namespace-docs.R
```diff
@@ -1967,6 +1967,40 @@ NULL
 #' @export
 NULL
 
+#' fftfreq
+#'
+#' Computes the discrete Fourier Transform sample frequencies for a signal of size `n`.
+#'
+#' @note 
+#' By convention, `fft()` returns positive frequency terms first, followed by the negative
+#' frequencies in reverse order, so that `f[-i]` for all `0 < i ≤n/2`
```
dfalbel, line 1976: The character `≤` here is probably causing an error in the CI:

```
❯ checking PDF version of manual ... [21s/21s] WARNING
  LaTeX errors when creating PDF version.
  This typically indicates Rd problems.
  LaTeX errors found:
  ! LaTeX Error: Unicode character ≤ (U+2264)
                 not set up for use with LaTeX.
```

R/gen-namespace-docs.R
```diff
@@ -1967,6 +1967,40 @@ NULL
...
+#' 
+#' @note 
+#' For even lengths, the Nyquist frequency at `f[n/2]` can be thought of as either negative
+#' or positive. `fftfreq()` follows NumPy’s convention of taking it to be negative.
+#'
+#' @param n (integer) – the FFT length
+#' @param d (float, optional) – the sampling length scale. The spacing between individual
+#' samples of the FFT input. The default assumes unit spacing, dividing that result by the
+#' actual spacing gives the result in physical frequency units.
+#' @param out (default: `NULL`) the output tensor
```
dfalbel, line 1989: You can't document an argument that's not in the function signature. In this case, the only way to match pytorch's signature is to write a wrapper function that adds those arguments to the signature. 

```
❯ checking Rd \usage sections ... WARNING
  Undocumented arguments in documentation object 'torch_fft_fftfreq'
    ‘options’
  Documented arguments not in \usage in documentation object 'torch_fft_fftfreq':
    ‘out’ ‘dtype’ ‘layout’ ‘device’ ‘requires_grad’
```

You can the function name to the list here: 

https://github.com/mlverse/torch/blob/main/tools/torchgen/R/r.R#L130

Regenerate the wraper code so the automatically wraped function name starts with a `.` then
You can write the wraper function in this file adding those arguments to the signature and mangling them
to create the correct call to `.torch_fft_fftreq`.

https://github.com/mlverse/torch/blob/main/R/wrapers.R

R/gen-namespace-docs.R
skeydan: Thanks, so great you're already pointing this out! (I hadn't checked out the exact reason yet, because I wanted to fix other things first)

skeydan:
Hi Daniel,
sorry, I _should_ have asked about the right workflow :-)

I've added `fft_fttfreq` to `r.R`, and tried sourcing and running that function, but that resulted in consecutive errors, and I thought I may need to use the `torchgen` project for that. The first time I did that, it didn't throw errors, but nothing was generated either.

I then thought that maybe I'd need to clean up `gen-namespace-docs` first (checking out that file from main), but now I get an error in `torchgen` (error opening `/src/gen-namespace.cpp`).

Maybe I should do a fresh checkout of `torch`? And could you please explain how to correctly do the generation?

thanks!

skeydan:
Hm, wanted to work on https://github.com/mlverse/torch/issues/935 in the meantime, but I already see that I will need to do the very same thing there :-)

So I'll better wait with that until I know how to correctly do it.

dfalbel:
To regenerate you can run:

```
devtools::load_all(); generate()
```

In the torchgen project.

skeydan:
Yes, that's what I did ... that's what I was describing, above:

> The first time I did that, it didn't throw errors, but nothing was generated either.

I then thought that maybe I'd need to clean up gen-namespace-docs first (checking out that file from main), but now I get an error in torchgen (error opening /src/gen-namespace.cpp).

skeydan:
Yeah, on a fresh branch it works!!

```
-#' @rdname torch_fft_fftfreq
-torch_fft_fftfreq <- function(n, d = 1L, options = list()) {
+#' @rdname .torch_fft_fftfreq
+.torch_fft_fftfreq <- function(n, d = 1L, options = list()) {

```

I'll close this PR, and link it (for information) in the new one.

PR #945, opened 2022-12-13
docs: fix nn_glu class name

sebffischer:


dfalbel:
Thanks very much @sebffischer !

PR #950, opened 2022-12-19
Add fftfreq (open question(s))

skeydan:
Hi @dfalbel, here's my next try at `fftfreq()` :-).

I still have two questions (one of which is a problem, the other one just something to clarify at this point).

The problem: I see that I'm handling the `out` tensor incorrectly (this is evident from the test in `wrapers.R` already).
I was looking for examples for how to do this in the code base, but didn't find anything that matched closely enough, so what I did was following what is seen in `test-storage.R` ... Probably this could not even work in this case, since here, a function call is involved ...
In reality, I do not even see why that option would be useful (to be able to pass in a tensor here), so if you like we could also just remove that argument ... (I would still be interested in knowing whether it'd be feasible in principle).

The question: How does `gen-namespace-examples.R` work? I see it used also for function which, in `gen-namespace-docs.R`, already have examples? I guess I don't need to add anything there, but if I did, how'd I generate the hash?

Thanks!!


skeydan:
Oh wait ... (re "problem") ... started looking at another PR-to-be (for `torch_where`) and realized that what I need is a complementing `_out` function! And that exists already, and as far as I see, I can just leave it as-is ... since I wouldn't know why someone would want to export it! Right?

Well in fact there still is one thing I'll modify there: the `d` argument should not have a default of `1L`, since it's a float (nothing to do with dimensions) ... I think it should be fine to make that change inside this PR, since I'm making the analogous modification in `fftfreq` ... 

Going to update the PR, please let me know if you disagree :-)

skeydan:
Me again... I think what I proposed above isn't the correct way to handle the `1L`  ... (shouldn't update `gen-namespace.R` manually anyway) ... I see this in `r.R`:

```
if (default %in% c("1", "0", "-1", "2", "-2", "100", "-100",
                     "20")) {
    return(paste0(default, "L"))
  }

```

What do we do about this? I'll leave it untouched for now, and just remove the case-handling code for `out` as I said before.

skeydan:
OK, I think that should do it ...

R/gen-namespace-docs.R
```diff
@@ -1967,6 +1967,40 @@ NULL
...
+#' 
+#' @note 
+#' For even lengths, the Nyquist frequency at `f[n/2]` can be thought of as either negative
+#' or positive. `fftfreq()` follows NumPy’s convention of taking it to be negative.
+#'
+#' @param n (integer) – the FFT length
+#' @param d (float, optional) – the sampling length scale. The spacing between individual
+#' samples of the FFT input. The default assumes unit spacing, dividing that result by the
+#' actual spacing gives the result in physical frequency units.
+#' @param dtype (default: `torch_get_default_dtype()`) the desired data type of returned tensor.
```
dfalbel, line 1989:
```suggestion
#' @param dtype (default: [torch_get_default_dtype()]) the desired data type of returned tensor. 
```

R/gen-namespace-docs.R
```diff
@@ -1967,6 +1967,40 @@ NULL
...
+#' @note 
+#' For even lengths, the Nyquist frequency at `f[n/2]` can be thought of as either negative
+#' or positive. `fftfreq()` follows NumPy’s convention of taking it to be negative.
+#'
+#' @param n (integer) – the FFT length
+#' @param d (float, optional) – the sampling length scale. The spacing between individual
+#' samples of the FFT input. The default assumes unit spacing, dividing that result by the
+#' actual spacing gives the result in physical frequency units.
+#' @param dtype (default: `torch_get_default_dtype()`) the desired data type of returned tensor. 
+#' @param layout (default: `torch_strided()`) the desired layout of returned tensor.
```
dfalbel, line 1990:
```suggestion
#' @param layout (default: [torch_strided()]) the desired layout of returned tensor. 
```

R/gen-namespace-docs.R
```diff
@@ -1967,6 +1967,40 @@ NULL
 #' @export
 NULL
 
+#' fftfreq
+#'
+#' Computes the discrete Fourier Transform sample frequencies for a signal of size `n`.
+#'
+#' @note 
+#' By convention, `fft()` returns positive frequency terms first, followed by the negative
```
dfalbel, line 1975:
```suggestion
#' By convention, [torch_fft_fft()] returns positive frequency terms first, followed by the negative
```

dfalbel (approved):
Looks great! Just minor updates in the docs.

dfalbel:
Regarding the defaults: since you already use the correct default in the wrapper function there's no need to modify the code generation or the generated function as it will never be called with its own default.

Regarding `out`: I think most of this `out` arguments are used for saving memory, by writing the results of a computation in an unused buffer, but we don't currently support those in R and if add support for them in the future it will probably be made in the C/C++ level instead.

Regarding examples: There's some functionality in [here](https://github.com/mlverse/torch/blob/main/tools/torchgen/R/docs.R) that was used to generate the docs but I would be surprised if it still works. It has been a long time I don't try it - it was mostly used to generate the first batch of functions.

skeydan:
Many thanks for the review and the explanations, - got it!

PR #951, opened 2022-12-19
#935

### LINKED ISSUE #935: single-argument torch.where returns 0-based indices

skeydan:
Fix #935


R/wrapers.R
```diff
@@ -472,6 +472,16 @@ torch_nonzero <- function(self, as_list = FALSE) {
   }
 }
 
+#' @rdname torch_where
+torch_where <- function(condition, self = NULL, other = NULL) {
+  if (is.null(self) && is.null(other)) {
+    out <- torch_nonzero(condition, as_list = TRUE)
```
dfalbel, line 478: I don't think it's a good idea to use a different operation here, unless this is how it's implemented in Python too. This would make it harder to debug in cases like tracing where only the operator name is recorded.

skeydan: I see, thanks! Pushing the updated version in a sec.
However, 

unless this is how it's implemented in Python too. 

https://github.com/pytorch/pytorch/blob/c780610f2d8358297cb4e4460692d496e124d64d/aten/src/ATen/native/TensorCompare.cpp#L132

I guess you could say it is, in a way (although that's not Python ;-))

Issue #955, opened 2023-01-09
Warning in R-devel


dfalbel:
We are getting:

```
checking compiled code ... WARNING
File ‘torch/libs/torch.so’:
  Found ‘sprintf’, possibly from ‘sprintf’ (C)
    Object: ‘xptr.o’
File ‘torch/libs/torchpkg.so’:
  Found ‘sprintf’, possibly from ‘sprintf’ (C)
    Object: ‘xptr.o’

Compiled code should not call entry points which might terminate R nor
write to stdout/stderr instead of to the console, nor use Fortran I/O
nor system RNGs nor [v]sprintf.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual.
```

This should be fixed until January 23 on CRAN to avoid the package removal.

shikokuchuo:
Provided a patch in PR #956 as I had to deal with the exact same issue recently in 'nanonext'.

dfalbel:
Thanks very much @shikokuchuo !

PR #956, opened 2023-01-09
Update xptr.cpp addresses issue #955

shikokuchuo:
Use snprintf instead of printf for added safety; remove spurious casts


dfalbel:
Thank you very much!

Issue #971, opened 2023-01-27
torch 0.9.0 and 0.9.1 deadlock when run in parallel


jonlachmann:
Hi,

Pytorch/libtorch as of version 1.12 allows forking in combination with autograd when only using the CPU and not the GPU as in this issue here https://github.com/pytorch/pytorch/issues/69839.

Previously (up until 0.8.1), running torch from R inside mclapply through this package was possible when setting ```torch_set_num_threads(1)```. Now, a deadlock situation emerges when doing so.

I have created a minimal example (similar to as in a previous issue I created) the code is:

```
library(torch)
library(parallel)

testfun <- function (x) {
  lstm <- nn_lstm(50, 50)
  in_data <- torch_tensor(matrix(rnorm(50),1), torch_float())
  out_data <- torch_tensor(array(rnorm(2500), c(1,50,50)), torch_float())
  out_pred <- lstm(in_data)[[1]]
  loss <- nnf_mse_loss(out_pred, out_data)
  loss$backward()
}

mclapply(1:2, testfun)
```

When running this code, a number of threads are spawned, and a deadlock situation where the CPU usage is 0% and the code never completes is created.

The thread tree in htop looks like this:
![Screenshot 2023-01-27 at 11 33 27](https://user-images.githubusercontent.com/10993608/215068786-5508ad1e-a21a-49d9-b945-28497e56c0d2.png)

Investigating the threads in gdb we can see that there seems to be a lockup in the ```std::condition_variable``` used in lantern/utils.cpp. https://github.com/mlverse/torch/blob/8369336cc25df518f6791917ad2653661b1f1a28/lantern/src/utils.cpp#L325

We see two threads stuck here:
![cond_var1](https://user-images.githubusercontent.com/10993608/215069812-967ba99b-f0ec-4732-9957-20a3aa3784b6.png)
 
5 threads are stuck here: 
![cond_var3](https://user-images.githubusercontent.com/10993608/215069915-7973fdd5-af05-4742-b493-827d3d0319ef.png)

And 5 threads stuck in this location:
 ![cond_var9](https://user-images.githubusercontent.com/10993608/215070010-3795d5c4-b6db-46af-a97d-ab308d0bf29d.png)

There is also one thread in select.c:
![select](https://user-images.githubusercontent.com/10993608/215070111-608d039a-ea0f-4947-b02d-9aaea6855517.png)

One in nanosleep:
[...]

PR #979, opened 2023-02-10
Only initialize the thread pool when backward is called for the first time.

dfalbel:
This is a step to unlock #971


tests/testthat/test-fork.R
```diff
@@ -0,0 +1,20 @@
+test_that("Forking doesn't dealock", {
```
skeydan, line 1:
```suggestion
test_that("Forking doesn't deadlock", {
```

dfalbel: Thx!

PR #1023, opened 2023-04-25
Wrap access to JIT operators

skeydan:


dfalbel:
I think you need to rebase on main so tests works correctly. Currently it's trying to download the wrong pre-built lanern libraries.

tests/testthat/test-ops.R
```diff
@@ -0,0 +1,23 @@
+test_that("can access operators via ops object", {
+  # matmul, default use
+  res <- ops$aten$matmul(torch::torch_ones(5, 4), torch::torch_rand(4, 5))
+  expect_equal(length(res), 1)
```
dfalbel, line 4: Not necessarily in this PR, but I wonder if given the return type of the schema that was used, if it would be possible to 'unlist' the output. For instance, we know that `ops$aten$matmul` returns a `TensorType` (from the schema). Could we in this case instead of returning a list, return the Tensor directly?

R/ops.R
```diff
@@ -0,0 +1,44 @@
+#' @export
+ops <- structure(list(), class = "torch_ops")
```
dfalbel, line 2: I like the name `ops` but it might cause issues with namespacing. What do you think about renaming this object to `jit_ops`?

dfalbel (commented):
Looks great @skeydan !

R/ops.R
skeydan: Yeah, I also thought `ops` might be too general ... `jit_ops` is great!

tests/testthat/test-ops.R
skeydan: Oh yes, that makes a lot of sense! Will do in this one ...

skeydan:
Done like we said and added a bunch of new tests :-)
I've noticed one other thing though, which I'm not sure is problematic or not ...

When you call on operator through `jit_ops`, you lose a lot of convenience functionality you're accustomed to from the R
wrappers - compare

```
res_torch <- torch_split(torch::torch_arange(0, 3), 4, 1)
res_jit <- jit_ops$aten$split(torch::torch_arange(0, 3), jit_scalar(4L), jit_scalar(0L))
```

1) You have to pass default arguments (which would be the case in Python too, I assume)
2) You have to use `jit_scalar` for scalar types (which I guess is fine)
3) You don't get the R conventions like axis numbering starting from 1, etc.)

Thinking about how this will be used, it might be ok though - what do you think @dfalbel?

R/jit-ops.R
```diff
@@ -0,0 +1,45 @@
+#' @export
```
dfalbel, line 1: Last thing! I think we need to add at least some documentation to that object. Basically a title, description and a few examples on usage would be nice. Also add it to the pkgdown index in: https://github.com/mlverse/torch/blob/main/_pkgdown.yml

dfalbel:
Cool! It's nice that it works by just flattenting the list structure.
Is there an example of a function where the number of returns is zero? what happens in that case?

I think you are right, it's up to library wrappers to implement scalar cast, etc. We won't have enough context to automate this for every extension.

R/jit-ops.R
skeydan: done :-)
As to the pkgdown index, I think it should already get included (`starts_with("jit")`), right?

Issue #1037, opened 2023-05-11
LBFGS w/ line search: Error in if (min(max(bracket) - t, t - min(bracket)) < eps)


skeydan:
```
library(luz)
library(torch)

net <- nn_module(
  "Net",
  initialize = function(num_hidden) {
    self$fc1 <- nn_linear(1, num_hidden)
    self$fc2 <- nn_linear(num_hidden, 1)
  },
  forward = function(x) {
    x <- self$fc1(x)
    x <- nnf_relu(x)
    x <- self$fc2(x)
    x
  },
  step = function() {
    if (ctx$training) {
      closure <- function() {
        pred <- ctx$model(ctx$input)
        loss <- ctx$loss_fn(pred, ctx$target)
        loss$backward()
        loss
      }
      ctx$loss <- ctx$opt$step(closure)
    } else {
      pred <- ctx$model(ctx$input)
      ctx$loss <- ctx$loss_fn(pred, ctx$target)
    }
  }
)

train_data <- list(torch_arange(1, 100)$unsqueeze(2), 2 * torch_arange(1, 100)$unsqueeze(2))

fitted <- net %>% 
  setup(
    loss = nn_mse_loss(),
    optimizer = optim_lbfgs
  ) %>% 
  set_hparams(num_hidden = 16) %>%
  set_opt_hparams(line_search_fn = "strong_wolfe") %>%
  fit(train_data, epochs = 5, valid_data = train_data)

```

```
Epoch 1/5
Train metrics: Loss: 5625.7354                                                 
Valid metrics: Loss: 56.826
Epoch 2/5
Train metrics: Loss: 17.7523                                                   
Valid metrics: Loss: 0.2049
Epoch 3/5
Train metrics: Loss: 0.1796                                                    
Valid metrics: Loss: 0.1856
Epoch 4/5
Error in if (min(max(bracket) - t, t - min(bracket)) < eps) { :
missing value where TRUE/FALSE needed
```

PR #1048, opened 2023-05-18
remove interim conversions to R

skeydan:
Fixes https://github.com/mlverse/torch/issues/1037.

The actual error depends on random state; it neither always appears nor originates from the same source location.
I was able to reliably produce one of its incarnations modifying the code from the issue like this:

```
set.seed(777)
torch_manual_seed(777)
train_data <- list(torch_arange(1, 100)$unsqueeze(2), 2 * torch_arange(1, 100)$unsqueeze(2))
```

At this point, it looks like just removing the interim conversions to R is enough; but I'll try further initializations.
At first, it seemed like `torch_nan_to_num()` would be a candidate to solve this, but right now it seems the `nan`s per se aren't the problem.


skeydan:
Added a test, and did some further tests sampling random values. I think it should be fine now, and ready for review.

PR #1075, opened 2023-06-28
better handling of partial argument matching

sebffischer:
In the CI of mlr3torch I noticed the partial argument matching of `con` to `connection` in `serialize()`.
This is fixed in this PR and additionally, I added some option setting during the tests to warn when something like this happens (let's see what comes out in the CI or whether you think this is a valuable setting).


sebffischer:
It seems like these failures are not related to the PR.

dfalbel (approved):
Thank you very much @sebffischer !

PR #1081, opened 2023-07-14
fix: device printer

sebffischer:


dfalbel:
Thanks very much @sebffischer !

PR #1087, opened 2023-08-10
docs: correct wrong information about indexing

sebffischer:


sebffischer:
The docs also show it, but here is a similar example as well: 

``` r
library(torch)
#> Warning: package 'torch' was built under R version 4.3.0

x = torch_ones(10)

x$reshape(c(2, 5))
#> torch_tensor
#>  1  1  1  1  1
#>  1  1  1  1  1
#> [ CPUFloatType{2,5} ]

x[1]
#> torch_tensor
#> 1
#> [ CPUFloatType{} ]
```

<sup>Created on 2023-08-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #1097, opened 2023-09-06
Solved `out_features` typo in `nn_linear`.

nachodieez:
The public member of `nn_linear` for the output features should be called `out_features` instead of `out_feature` (s was missing). Solved the typo and updated `test-nn.R` to not fail with the new name.


dfalbel:
Thanks very much @nachodieez 
I'll merge as soon as tests run!

PR #1106, opened 2023-09-26
add FR messages translation

cregouby:
Provide the french translation of all messages.
The R translation framework relies on message detection through `gettext()` (out of the `base::stop()` `message()` and `warnings()`) thus I've wrapped all messages through it. 
I've tried to add the `gettext()`wrapper inside each the message definition in `conditions.R` file. This would have made the code nicer but after this, messages are not detected through the `potools::po_extract(".")` tolls to aggregate the translation file, wich means they won't be translated.
This makes the P.R. looks like massive. Sorry for this
Few messages typo are fixed as well


dfalbel:
Hi @cregouby,

It's very unfortunate that `potools` can't find translations without us having to wrap all messages with `gettext`. Do you think it's worth opening a feature request for them? It seems they have [support for `messagef`, `warningf` variants](https://michaelchirico.github.io/potools/articles/developers.html#extraction) and maybe it's possible for them to add support for custom 'message raising' functions.

I don't want to block the PR tough, it looks great to me and it's very nice to have the translated messages!

Can you please add a small section to the [CONTRIBUTING](https://github.com/mlverse/torch/blob/main/.github/CONTRIBUTING.md) file with some instructions to add new translations and how to check them?

Best,
Daniel

cregouby:
Hello @dfalbel,
Thanks for the review.
I'll both raise a feature request to {potools} team and update the CONTRIBUTING file.
And I'll let you know. 

Best,
Christophe

PR #1121, opened 2023-11-13
exports torch_serialize()

shikokuchuo:
I am assuming this function is meant to be exported as it is in the docs and was announced as part of the v0.9.0 updates.

Discovered when I was refining the serialisation hook functionality in `nanonext` (https://github.com/shikokuchuo/nanonext/discussions/15) , testing against torch serialisation versions 2 and 3 .


dfalbel:
You are correct @shikokuchuo !
Thanks for catching this and for the PR :)

PR #1127, opened 2023-12-26
Added information about m1 GPU

cyrusmsk:
Based on messages from issues, it seems that GPU is supported on Apple Silicon through MPS (Metal Performance Shaders). Same approach is used in PyTorch.

I think it should be presented in the documentation.


dfalbel:
Thank you!

Issue #1130, opened 2024-01-25
Question about NEWS


sebffischer:
The news mentions, that tensors are now moved to cpu in `as.array(), but there is still this line: https://github.com/mlverse/torch/blob/92a8a05a3d4963f1d740939de087ad35a822ccf9/R/tensor.R#L415

Is this intentional?

dfalbel:
Ohhh this not intentional at all. We should remove that so we don't raise the error any longer.

PR #1132, opened 2024-01-26
fix(tensor): cuda to r conversion

sebffischer:
Addresses https://github.com/mlverse/torch/issues/1130


sebffischer:
Thanks a lot! I did not catch the test failures as I was working on a Mac …

dfalbel:
No worries @sebffischer, thank you for the PR. The CI pipelines were broken for the GPU's too, I only recovered them recently!

Issue #1126, opened 2023-12-18
Cloning of torch module behaves unexpectedly


sebffischer:
Thanks a lot already for fixing the class of the cloned torch module! 
I still observed some differences between the cloned object and the clonee, see below where the `weight` of the linear layer is missing the "nn_parameter" class.

``` r
library(mlr3torch)
#> Loading required package: mlr3
#> Loading required package: mlr3pipelines
#> Loading required package: torch

a = nn_linear(1, 1)

b = a$clone(deep = TRUE)

a
#> An `nn_module` containing 2 parameters.
#> 
#> ── Parameters ──────────────────────────────────────────────────────────────────
#> • weight: Float [1:1, 1:1]
#> • bias: Float [1:1]

b
#> An `nn_module` containing 2 parameters.
#> 
#> ── Parameters ──────────────────────────────────────────────────────────────────
#> • weight: Float [1:1, 1:1]
#> • bias: Float [1:1]

b$parameters$weight |> attributes()
#> $class
#> [1] "torch_tensor" "R7"

a$parameters$weight |> attributes()
#> $class
#> [1] "torch_tensor" "R7"           "nn_parameter"
```

<sup>Created on 2023-12-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
I tried to do this myself, but because `clone` is autogenerated, I had to change the Declarations file (I believe), but when I try to autogenerate the functions I get: 

```r
> torchgen::generate("~/gh/torch")
Starting code generation ...
Error in `purrr::map_chr()`:
ℹ In index: 1.
Caused by error:
! Result must be length 1, not 0.
```

sebffischer:
This kind of still breaks when one wants to do something like `nn_parameter(torch_tensor(1))$clone()`.
While admittedly this might be less important, I think the (autogenerated)`torch_clone()` method itself should just preserve the attributes.

dfalbel:
I see, you are right! Perhaps you can add a method like `clone_with_attributes` around here:

https://github.com/mlverse/torch/blob/0e9fdd78852601b655eed48a2eaf5e22033dead0/R/tensor.R#L14

With the behavior you desire and then I can make the plumbing for it to be the default `clone`. 

TBH I haven't executed much torchgen like this, almost always doing load_all() then `generate()` with the defaults. So maybe there's a hardcode path that doesn't work properly when running from a different directory.

PR #1134, opened 2024-02-07
fix(Tensor): cloning preserves attributes

sebffischer:
Addresses https://github.com/mlverse/torch/issues/1126, where you offered to take care of the renaming. 
I.e. the `$clone2()` method should just be renamed to `$clone()`, which currently has no effect, as the `find_method()` function (iirc) still finds the `torch_clone()` function that is auto-generated.

### DIFF
--- a/R/tensor.R
+++ b/R/tensor.R
@@ -71,27 +71,27 @@ Tensor <- R7Class(
     },
     to = function(dtype = NULL, device = NULL, other = NULL, non_blocking = FALSE,
                   copy = FALSE, memory_format = NULL) {
-      
+
       has_device <- !is.null(device)
       has_dtype <- !is.null(dtype)
       has_other <- !is.null(other)
-      
+
       if (has_other) {
         # can't have device and dtype
         if (has_device || has_dtype) {
           cli::cli_abort("Had {.arg other} but {.arg device} or {.arg dtype} are non {.val NULL}")
         }
-        
+
         return(private$`_to`(other = other, non_blocking = non_blocking, copy = copy))
       }
-      
+
       if (!has_dtype) {
         dtype <- self$dtype
       }
-      
+
       if (has_device) {
         private$`_to`(
-          dtype = dtype, 
+          dtype = dtype,
           device = device,
           non_blocking = non_blocking,
           copy = copy,
@@ -288,6 +288,11 @@ Tensor <- R7Class(
     },
     half = function() {
       self$to(dtype=torch_half())
+    },
+    clone2 = function() {
+      x <- torch_clone(self)
+      attributes(x) <- attributes(self)
+      return(x)
     }
   ),
   active = list(
@@ -383,7 +388,7 @@ as.matrix.torch_tensor <- function(x, ...) {
 as_array_impl <- function(x) {
   # move tensor to cpu before copying to R
   x <- x$cpu()
-  
+
   if (x$is_complex()) {
     out <- complex(
       real = as.array(x$real),
@@ -392,7 +397,7 @@ as_array_impl <- function(x) {
     dim(out) <- dim(x)
     return(out)
   }
-  
+
   a <- cpp_as_array(x)
 
   if (length(a$dim) <= 1L) {
@@ -479,13 +484,13 @@ Tensor$set("active", "ptr", function() {
 
 tensor_to_complex <- function(x) {
   torch_complex(
-    torch_tensor(Re(x), dtype = torch_double()), 
+    torch_tensor(Re(x), dtype = torch_double()),
     torch_tensor(Im(x), dtype = torch_double())
   )
 }
 
 #' Creates a tensor from a buffer of memory
-#' 
+#'
 #' It creates a tensor without taking ownership of the memory it points to.
 #' You must call `clone` if you want to copy the memory over a new tensor.
 #'
--- a/tests/testthat/test-tensor.R
+++ b/tests/testthat/test-tensor.R
@@ -408,14 +408,14 @@ test_that("print complex tensors", {
   skip_on_os("linux")
   x <- torch_complex(torch_randn(10), torch_randn(10))
   expect_snapshot(
-    print(x)  
+    print(x)
   )
-  
+
   x$requires_grad_(TRUE)
   expect_snapshot(
     print(x)
   )
-  
+
   y <- 2*x
   expect_snapshot(
     print(y)
@@ -423,52 +423,52 @@ test_that("print complex tensors", {
 })
 
 test_that("complex tensors modifications and acessing", {
-  
+
   real <- torch_randn(10, 10)
   imag <- torch_randn(10, 10)
   x <- torch_complex(real, imag)
-  
+
   expect_equal_to_tensor(x$real, real)
   expect_equal_to_tensor(x$imag, imag)
-  
+
   real <- torch_randn(10, 10)
   imag <- torch_randn(10, 10)
   x$real <- real
   x$imag <- imag
-  
+
   expect_equal_to_tensor(x$real, real)
   expect_equal_to_tensor(x$imag, imag)
-  
+
 })
 
 test_that("create complex from and to R", {
-  
+
   x <- complex(real = c(0, 1), imaginary = c(1, 1))
   y <- torch_tensor(x)
-  
+
   expect_equal_to_r(y$imag, Im(x))
   expect_equal_to_r(y$real, Re(x))
   expect_true(y$dtype == torch_cfloat())
-  
+
   x <- complex(real = runif(1), imaginary = runif(1))
   y <- torch_tensor(x, dtype = torch_cdouble())
-  
+
   expect_equal_to_r(y$imag, Im(x))
   expect_equal_to_r(y$real, Re(x))
   expect_true(y$dtype == torch_cdouble())
-  
+
   x <- torch_complex(torch_randn(10, 10), torch_randn(10, 10))
   y <- as.array(x)
   z <- torch_tensor(y)
-  
+
   expect_true(torch_allclose(x, z))
-  
+
   x <- torch_complex(1, 1)
   y <- as.array(x)
   z <- torch_tensor(y)
   expect_true(torch_allclose(x, z))
   expect_equal(as.complex(x), complex(real = 1,imaginary = 1))
-  
+
 })
 
 test_that("expand works", {
@@ -496,22 +496,22 @@ test_that("is_sparse works", {
 })
 
 test_that("can make a byte tensor from a raw vector", {
-  
+
   x <- charToRaw("hello world")
   ten <- torch_tensor(x)
-  
+
   expect_equal(ten$dtype$.type(), "Byte")
   expect_equal(length(x), length(ten))
-  
+
   expect_equal(as.array(ten), x)
   expect_equal(rawToChar(as.array(ten)), "hello world")
 })
 
 test_that("to can change both device and dtype", {
-  
+
   x <- torch_randn(10, 10)
   y <- x$to(dtype = "double", device = "meta")
-  
+
   expect_true(y$dtype == torch_double())
   expect_true(y$device == torch_device("meta"))
 })
@@ -546,6 +546,28 @@ test_that("can copy a mps tensor", {
   x <- array(runif(100), dim = c(10, 10))
   y <- torch_tensor(x, device="mps")
   x_ <- as.array(y)
-  
+
   expect_true(all.equal(x, x_, tolerance = 1e-5))
+})
+
+test_that("cloning works and preserves attributes", {
+  # buffer
+  x_buf <- nn_buffer(torch_tensor(1))
+  x_buf_clone = x_buf$clone2()
+  expect_equal(attributes(x_buf), attributes(x_buf_clone))
+  x_buf[1] = 2
+  expect_false(torch_equal(x_buf, x_buf_clone))
+
+  x_param <- nn_parameter(torch_tensor(1))
+  x_param_clone = x_param$clone2()
+  expect_equal(attributes(x_param), attributes(x_param_clone))
+  x_param$requires_grad_(FALSE) # otherwise we cannot modify tensor in-place
+  x_param[1] = 2
+  expect_false(torch_equal(x_param, x_param_clone))
+
+  x <- torch_tensor(1)
+  x_clone <- x$clone2()
+  expect_equal(attributes(x_clone), attributes(x))
+  x[1] <- 2
+  expect_false(torch_equal(x_clone, x))
 })
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,6 +10,7 @@ Authors@R: c(
     person("Christophe", "Regouby", role = c("ctb")),
     person("Krzysztof", "Joachimiak", role = c("ctb")),
     person("Hamada S.", "Badr", role = c("ctb")),
+    person("Sebastian", "Fischer", role = c("ctb")),
     person(family = "RStudio", role = c("cph"))
     )
 Description: Provides functionality to define and train neural networks similar to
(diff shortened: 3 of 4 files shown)


sebffischer:
Also, the Cpp code to clone tensors / buffers / parameters behaves differently, right?

sebffischer:
I also think it is important here, that `torch_clone(tensor)` and `tensor$clone()` behave identically.

sebffischer:
There is also another discrepancy between the clonee and the cloned object: 

``` r
library(torch)
lin = nn_linear(1, 1)
lin$parameters$weight$requires_grad
#> [1] TRUE
lin$clone(deep = TRUE)$parameters$weight$requires_grad
#> [1] FALSE
```

<sup>Created on 2024-02-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
one more: 

``` r
library(torch)
nn_linear(1, 1)$train()$clone(deep = TRUE)
#> Error in FUN(X[[i]], ...): not an environment
nn_linear(1, 1)$eval()$clone(deep = TRUE)
#> Error in FUN(X[[i]], ...): not an environment
```

<sup>Created on 2024-02-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
Also, cloning repeatedly causes issues and builds up a structure of parent environments.
This is, because the `clone` that is being retrieved here https://github.com/mlverse/torch/blob/0e9fdd78852601b655eed48a2eaf5e22033dead0/R/nn.R#L521 when cloning the second time is already the patched version and not the original R6 clone implementation. 
The patched version still has the original version as its enclosing environment so it should still work but will repeatedly call the `clone()` method of the enclosing environment (https://github.com/mlverse/torch/blob/0e9fdd78852601b655eed48a2eaf5e22033dead0/R/nn.R#L549) until it reaches the top-level clone call, i.e. R6's clone method.

``` r
library(torch)

n = nn_linear(1, 1)
head(attr(n, "module")$clone, n = 1)
#>                                                       
#> 1 function (deep = FALSE, ..., replace_values = TRUE)
head(attr(n, "module")$clone |> environment() |> with(clone), n = 1)
#>                           
#> 1 function (deep = FALSE)

n1 = n$clone(deep = TRUE)
head(attr(n1, "module")$clone, n = 1)
#>                                                       
#> 1 function (deep = FALSE, ..., replace_values = TRUE)
head(attr(n1, "module")$clone |> environment() |> with(clone), n = 1)
#>                                                       
#> 1 function (deep = FALSE, ..., replace_values = TRUE)
head(attr(n1, "module")$clone |> environment() |> with(clone) |> environment() |> with(clone), n = 1)
#>                           
#> 1 function (deep = FALSE)

identical(
  attr(n1, "module")$clone |> environment(),
  attr(n1, "module")$clone |> environment() |> with(clone) |> environment()
)
#> [1] FALSE

attr(n1, "module")$clone |> environment() |> names()
#> [1] "clone"    "f"        "instance"
attr(n1, "module")$clone |> environment() |> with(clone) |> environment() |> names()
#> [1] "clone"    "f"        "instance"
```

<sup>Created on 2024-02-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
Also, I included support for a private clone finalizer method. In mlr3torch we need something like this, because we have an `nn_module` that has an R6 class containing modules, but these modules also need to be registered in the nn_module class. So the reference identity of these objects needs to be kept when cloning and the only solution I came up with is to allow for some hook after calling clone. Let me know what you think :)

sebffischer:
This might be the last one that I have to address, i.e. nested modules do not seem to be cloned

``` r
library(torch)
nn_test = nn_module("test", initialize = function() {
  self$l = nn_module_list(list(nn_linear(1, 1)))
  },
  forward = function(x) {
    self$l[[1]](x)
  }
)()

nn_test1 = nn_test$clone(deep = TRUE)
nn_test$l$modules[[2]]
#> An `nn_module` containing 2 parameters.
#> 
#> ── Parameters ──────────────────────────────────────────────────────────────────
#> • weight: Float [1:1, 1:1]
#> • bias: Float [1:1]
nn_test1$l$modules[[2]]
#> An `nn_module` containing 2 parameters.
#> 
#> ── Parameters ──────────────────────────────────────────────────────────────────
#> • weight: Float [1:1, 1:1]
#> • bias: Float [1:1]
identical(nn_test$l$modules[[2]], nn_test1$l$modules[[2]])
#> [1] TRUE
```

<sup>Created on 2024-02-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
I also just saw: https://github.com/r-lib/R6/pull/273, which would make the post_clone hook officially supported by R6.

sebffischer:
So there is at least one more issue shown in the regex below. When creating the `state_dict`, the parameters of the children are not being traversed here: https://github.com/mlverse/torch/blob/0e9fdd78852601b655eed48a2eaf5e22033dead0/R/nn.R#L543

I think we need to recurse through the children and things should probably work.

``` r
library(torch)
nn_test = nn_module("test", initialize = function() {
  self$l = nn_module_list(list(nn_linear(1, 1)))
  },
  forward = function(x) {
    self$l[[1]](x)
  }
)()

nn_test1 = nn_test$clone(deep = TRUE)
nn_test$clone(deep = TRUE)
#> An `nn_module` containing 2 parameters.
#> 
#> ── Modules ─────────────────────────────────────────────────────────────────────
#> • l: <nn_module_list> #2 parameters
l1 = nn_test$l$modules[[2]]
l2 = nn_test1$l$modules[[2]]
identical(l1, l2)
#> [1] TRUE
```

<sup>Created on 2024-02-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
The current workaround for the clone method caused the cloned object to reference the original object. 
It was caused by this line: https://github.com/mlverse/torch/blob/0e9fdd78852601b655eed48a2eaf5e22033dead0/R/nn.R#L523

This implied that the size of the cloned object was larger than the original object:

``` r
library(torch)

pryr::object_size(nn_relu())
#> 484.47 kB
pryr::object_size(nn_relu()$clone(deep = TRUE))
#> 492.42 kB
```

<sup>Created on 2024-02-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sebffischer:
@dfalbel I am done here and would love to get your feedback whether you think these changes make sense :)

R/nn.R
```diff
@@ -516,45 +521,82 @@ create_nn_module_callable <- function(instance) {
...
+
+  # as R6's clone method is quite restrictive, we here assign the original public $clone() method to the private
+  # field $.__clone_r6__ and create a new patched $clone() method
+  # This circumvents some restrictions in R6, see e.g. this discussion: https://github.com/r-lib/R6/issues/179
   rlang::env_binding_unlock(instance, "clone")
   on.exit({lockBinding("clone", instance)}, add = TRUE)
-  clone <- instance$clone
-  
+  on.exit({lockBinding(".__clone_r6__", instance$.__enclos_env__$private)}, add = TRUE)
+  instance$.__enclos_env__$private$.__clone_r6__ = instance$clone
```
sebffischer, line 536:
```suggestion
  instance$.__enclos_env__$private$.__clone_r6__ <- instance$clone
```

R/nn.R
```diff
@@ -516,45 +521,82 @@ create_nn_module_callable <- function(instance) {
...
+      }
+      # also need to append a clone of the modules to this list.
+      # child modules can be duplicated - and have the same name
+      # note that we store both the modules, as well as their parameters and buffers in the state_dict
+      children <- instance$children
+      if (!length(children)) {
+        return(state_dict)
+      }
+      for (child in children) {
+        state_dict = collect_state_dict(child, state_dict)
```
sebffischer, line 555:
```suggestion
        state_dict <- collect_state_dict(child, state_dict)
```

R/nn.R
```diff
@@ -516,45 +521,82 @@ create_nn_module_callable <- function(instance) {
...
+      }
+      state_dict = append(state_dict, rlang::set_names(children, map_chr(children, rlang::obj_address)))
```
sebffischer, line 557:
```suggestion
      state_dict <- append(state_dict, rlang::set_names(children, map_chr(children, rlang::obj_address)))
```

R/nn.R
```diff
@@ -516,45 +521,82 @@ create_nn_module_callable <- function(instance) {
...
     }
-    
-    cloned_instance <- clone(deep = deep)
-    
+    cloned_instance <- private$.__clone_r6__(deep = deep)
+
     if (deep && replace_values) {
-      cloned_instance$.replace_values_from_table(state_dict)  
+      cloned_instance$.replace_values_from_table(state_dict)
+      cloned_private = cloned_instance$.__enclos_env__$private
```
sebffischer, line 590:
```suggestion
      cloned_private <- cloned_instance$.__enclos_env__$private
```

tests/testthat/test-nn.R
```diff
@@ -817,5 +817,79 @@ test_that("can clone a module with no state dict", {
...
+  l2 <- nn_test1$l$modules[[2]]
+  expect_false(identical(
+    l1$parameters,
+    l2$parameters
+  ))
+  expect_false(identical(l1, l2))
+})
+
+test_that("non-persistent buffers are cloned", {
+  n = nn_identity()
```
sebffischer, line 891:
```suggestion
  n <- nn_identity()
```

tests/testthat/test-nn.R
```diff
@@ -817,5 +817,79 @@ test_that("can clone a module with no state dict", {
...
+  n$register_buffer("a", nn_buffer(torch_tensor(1)), persistent = FALSE)
+  n1 = n$clone(deep = TRUE)
```
sebffischer, line 893:
```suggestion
  n1 <- n$clone(deep = TRUE)
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+
   expect_true(all.equal(x, x_, tolerance = 1e-5))
-})
\ No newline at end of file
+})
+
+test_that("cloning works and preserves attributes", {
+  # buffer
+  x_buf <- nn_buffer(torch_tensor(1))
+  x_buf_clone = x_buf$clone2()
```
sebffischer, line 556:
```suggestion
  x_buf_clone <- x_buf$clone2()
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  expect_equal(attributes(x_buf), attributes(x_buf_clone))
+  x_buf[1] = 2
```
sebffischer, line 558:
```suggestion
  x_buf[1] <- 2
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  expect_false(torch_equal(x_buf, x_buf_clone))
+
+  # parameter
+  x_param <- nn_parameter(torch_tensor(1))
+  x_param_clone = x_param$clone2()
```
sebffischer, line 563:
```suggestion
  x_param_clone <- x_param$clone2()
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  expect_equal(attributes(x_param), attributes(x_param_clone))
+  x_param$requires_grad_(FALSE) # otherwise we cannot modify tensor in-place
+  x_param[1] = 2
```
sebffischer, line 566:
```suggestion
  x_param[1] <- 2
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  # tensor
+  x <- torch_tensor(1)
+  x_clone <- x$clone2()
+  expect_equal(attributes(x_clone), attributes(x))
+  x[1] <- 2
+  expect_false(torch_equal(x_clone, x))
+})
+
+test_that("requires grad is left unchanged when cloning tensor", {
+  x_requires_grad = torch_tensor(1, requires_grad = TRUE)
```
sebffischer, line 578:
```suggestion
  x_requires_grad = torch_tensor(1, requires_grad <- TRUE)
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  x_no_requires_grad = torch_tensor(1, requires_grad = FALSE)
+  expect_true(x_requires_grad$clone()$requires_grad)
+  expect_false(x_no_requires_grad$clone()$requires_grad)
+})
+
+test_that("grad_fn and cloning", {
+  # this is the same behaviour as shown by PyTorch's .clone() method
+  x = torch_tensor(1, requires_grad = TRUE)
```
sebffischer, line 586:
```suggestion
  x <- torch_tensor(1, requires_grad = TRUE)
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  x1 = x$clone2()
```
sebffischer, line 587:
```suggestion
  x1 <- x$clone2()
```

tests/testthat/test-tensor.R
```diff
@@ -546,6 +546,44 @@ test_that("can copy a mps tensor", {
...
+  x <- torch_tensor(1)
+  x_clone <- x$clone2()
+  expect_equal(attributes(x_clone), attributes(x))
+  x[1] <- 2
+  expect_false(torch_equal(x_clone, x))
+})
+
+test_that("requires grad is left unchanged when cloning tensor", {
+  x_requires_grad = torch_tensor(1, requires_grad = TRUE)
+  x_no_requires_grad = torch_tensor(1, requires_grad = FALSE)
```
sebffischer, line 579:
```suggestion
  x_no_requires_grad <- torch_tensor(1, requires_grad = FALSE)
```

sebffischer:
Ok, now I think it is actually ready from my side

dfalbel:
@sebffischer I wonder if we instead of renaming `clone2` to `clone`, we could find a different name that clearly states what it does. The torch documentation clearly states:

> This function is differentiable, so gradients will flow back from the result of this operation to input. To create a tensor without an autograd relationship to input see [detach()](https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html#torch.Tensor.detach).

Thus renaming these methods will certainly cause problems in other codebases that rely on that behavior. For instance, it breaks some torch optimizers that do use `clone()` for that.

Besides that, the PR looks great! Thank you very much for working on this.

R/tensor.R
```diff
@@ -288,6 +288,13 @@ Tensor <- R7Class(
     },
     half = function() {
       self$to(dtype=torch_half())
+    },
+    clone2 = function() {
+      x <- torch_clone(self)
+      x$requires_grad_(self$requires_grad)
```
dfalbel, line 294: Can we add a comment on why we need to explicitly set `requires_grad` here?

dfalbel:
Ok, so renaming works fine, we just had to support the other arguments to `clone()` that were not specified.
The only thing I'm not convinced is that we want to set `requires_grad`, do you know why exactly we need this? Is it not respected?

sebffischer:
@dfalbel Thanks for the feedback. Regarding the call to `$requires_grad_()`: think I mixed something up. Indeed, the `$requires_grad` field does not need to be modified manually in the tensor's `$clone()` method.

sebffischer:
please don't merge yet I want to add one more refactor

sebffischer:
Done now

dfalbel:
Thanks @sebffischer ! Looks great!
I also ran luz and minhub tests and they all passed against this version.

PR #1135, opened 2024-02-09
feat: Added `compare_proxy` method for `torch_tensor`

sebffischer:
I left out the grad_fn because it e.g. changes when cloned, but a tensor should be equal to its cloned version


sebffischer:
I opened another PR

PR #1139, opened 2024-02-19
feat: comparison of torch tensors

sebffischer:
I did not create a PR in waldo, because I think this method should be under full control of the torch package. I "hacked" it in, so waldo does not become a hard dependency here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # torch (development version)
 
 - Make sure deep cloning preserve state dict attributes. (#1129)
+- A `compare_proxy` method for the `torch_tensor` type was added
+  it allows to compare torch tensor's using `testthat::expect_equal()`.
 
 # torch 0.12.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Make sure deep cloning preserve state dict attributes. (#1129)
 - A `compare_proxy` method for the `torch_tensor` type was added
-  it allows to compare torch tensor's using `testthat::expect_equal()`.
+  it allows to compare torch tensors using `testthat::expect_equal()`.
 - Converting torch tensor to R array works when tensor has 'cuda' device (#1130)
 
 # torch 0.12.0
--- a/R/compare.R
+++ b/R/compare.R
@@ -1,0 +1,34 @@
+compare_proxy.torch_tensor <- function(x, path) {
+  # we don't include the grad_fn because it should change after cloning a tensor (BackwardClone)
+  list(
+    object = list(
+      x = torch::as_array(x),
+      device = as.character(x$device),
+      requires_grad = x$requires_grad,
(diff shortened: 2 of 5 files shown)


R/compare.R
```diff
@@ -0,0 +1,34 @@
+compare_proxy.torch_tensor <- function(x, path) {
+  # we don't include the grad_fn because it should change after cloning a tensor (BackwardClone)
```
sebffischer, line 2:
```suggestion
```

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # torch (development version)
 
 - Make sure deep cloning preserve state dict attributes. (#1129)
+- A `compare_proxy` method for the `torch_tensor` type was added
+  it allows to compare torch tensor's using `testthat::expect_equal()`.
```
sebffischer, line 5:
```suggestion
  it allows to compare torch tensors using `testthat::expect_equal()`.
```

sebffischer:
@dfalbel what is your opinion on this, do you want this in the torch package at all?

dfalbel:
Yes, I think that's a useful feature! Do you mind checking why it fails on CI? We could also instead of registering onm package load, only register it when testing. Eg defining it in https://github.com/mlverse/torch/blob/main/tests/testthat/setup.R

sebffischer:
It should be fixed now. I would prefer to add it during package so other people (including myself) can use it in their tests as well.

sebffischer:
I am a bit confused by the failing windows test

dfalbel:
Yes, it's failing on older R versions I think:

```
Error: Error: argument "path" is missing, with no default
Backtrace:
     x
  1. +-testthat::expect_failure(expect_equal(x, x$clone())) at test-compare.R:37:3
  2. | \-testthat::capture_expectation(expr)
  3. |   \-base::tryCatch(...)
  4. |     \-base (local) tryCatchList(expr, classes, parentenv, handlers)
  5. |       \-base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
  6. |         \-base (local) doTryCatch(return(expr), name, parentenv, handler)
  7. \-testthat::expect_equal(x, x$clone())
  8.   \-testthat:::expect_waldo_equal("equal", act, exp, info, ..., tolerance = tolerance)
  9.     \-testthat:::waldo_compare(...)
 10.       \-waldo::compare(x, y, ..., x_arg = x_arg, y_arg = y_arg)
 11.         \-waldo:::compare_structure(x, y, paths = c(x_arg, y_arg), opts = opts)
 12.           +-waldo::compare_proxy(x)
 13.           \-torch:::compare_proxy.torch_tensor(x)
 ```

We could simply skip this tests if on R <= 4.0 if you want.

sebffischer:
Thx for fixing the typo! I think we are good to merge here?

dfalbel:
Thank you @sebffischer !

Issue #1136, opened 2024-02-12
Calling `$detach()` discards attributes


sebffischer:
``` r
library(torch)

x = nn_parameter(torch_randn(1))
class(x)
#> [1] "torch_tensor" "R7"           "nn_parameter"
class(x$detach())
#> [1] "torch_tensor" "R7"
```

<sup>Created on 2024-02-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dfalbel:
We could use a similar approach to what we did for clone, right?

sebffischer:
Yes, I will PR it. Just noted it down here so it won't be forgotten and I did not want to include even more changes in the cloning PR.

PR #1140, opened 2024-02-20
fix(tensor): detach preserves attributes

sebffischer:
Solves https://github.com/mlverse/torch/issues/1136


dfalbel:
Fantastic! Thank you very much @sebffischer !

PR #1141, opened 2024-02-21
Fix a few typos in messages, and align some variable decoration in messages

cregouby:
...as a preparation to the translation P.R.


cregouby:
You're welcome !

PR #1180, opened 2024-07-02
Use rlang instead of ellipsis for checking dots.

olivroy:
Since ellipsis imports rlang already and ellipsis is superseded. https://rlang.r-lib.org/news/#argument-intake-1-0-0


dfalbel:
Thanks @olivroy !

PR #1216, opened 2024-12-06
Fix/trace mode

sebffischer:
Bugfixes:
* save-load roundtrip used to convert `script_function` into `script_module`. This is fixed now (in a somewhat hacky way)

Open questions.
* In this PR, we only do the train-eval distinction for the `$forward` method. Maybe we want to do this for all methods.
* Can we register the `$forward` method (that distinguishes between the internal train and eval forward method) in the underlying script module instead of the `nn_Module`?
* Reading a script function is currently loaded as a module, I don't think this should be the case

TODOs:
* [ ] the training mode after saving a `script_module` is changed. This should probably be fixed.
* [ ] Currently, after saving a trace-jitted module, the `$forward` method is not defined. This is because the underlying module class is not saved when saving a trace jitted module, so when loading it again, the unpatched methods are used.
* [ ] Properly distinguish between saved module and saved function. The current hack does not work for jitted modules that were saved with an old version.


sebffischer:
Superseded by: https://github.com/mlverse/torch/pull/1217

PR #1217, opened 2024-12-10
respect train/eval mode in traced network

sebffischer:
Further issues:
* Reading a script function is currently loaded as a module, I don't think this should be the case
* Probably we should use a separate `CompilationUnit` per compiled module


.github/workflows/lantern.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
           # when changing supported versions here, please modify supported versions
           # in install.R
           - {os: macOS, version: cpu-intel, runner: macos-13}
-          - {os: macOS, version: cpu-m1, runner: [self-hosted, m1]}
+          - {os: macOS, version: cpu-m1, runner: macos-13}
```
dfalbel, line 22: I think you actually want `macos-14`, AFAICT macos-13 runs on intel macos runners. See the table in: https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images

sebffischer: thanks!

sebffischer:
Note that this maybe also fixes https://github.com/mlverse/torch/pull/633/files but I need to check again

sebffischer:
@dfalbel I think something is broken with the macOS runner. I don't think that this PR should behave differently on different operating systems, however.

dfalbel:
The runner was missing a `m1` label that I just added. It should run on m1 now for this PR

R/trace.R
```diff
@@ -295,20 +301,52 @@ jit_trace_module <- function(mod, ..., strict = TRUE) {
...
+        fn = tr_fn,
+        inputs = inp,
+        compilation_unit = .compilation_unit,
+        strict = strict,
+        module = module$..ptr..(),
+        name = "trainforward",
+        should_mangle = TRUE,
+        manage_memory = FALSE
+      )
+      mod$eval()
```
dfalbel, line 317: I wonder if we should add an argument to `jit_trace` that would keep the old behavior. My two concerns are:

1. Tracing runs the network twice, which could be problematic for some users.
2. Duplicates the size of the graph, which might be undesidered. Maybe a user wants to just trace the forward method in eval mode to export for deployment.

sebffischer: Ok, I changed my mind again, I think the default TRUE is fine, but I am happy to change it as well.
In which cases do you think running the network twice is problematic?

sebffischer: Added a `respect_mode` argument that triggers the double/single tracing

sebffischer:
@dfalbel Is this now okay?

dfalbel:
Do you know why the snapshots are different in the other OS'es? Is this something we can fix, or maybe just skip the test...

sebffischer:
@dfalbel I just assumes it's passing now. 
The problem is that to generate the unique names in the CompilationUnit a global state is maintained in `torch`.
Depending on how many other modules were registered in the compilation unit, the unique IDs change.
Because the other runners skip the M1 tests, they generate different unique names.

I think skipping is fine as long as they pass on one OS

dfalbel (commented):
LGTM!

Issue #1218, opened 2024-12-11
Segfault when running the tests twice


sebffischer:
When I run `devtools::test()` twice, I get

```r
> devtools::test()
ℹ Testing torch
✔ | F W  S  OK | Context
✔ |      1   0 | 0-threads                                                                                                                                                                                                                                                        
✔ |      1   2 | allocator                                                                                                                                                                                                                                                        
✔ |      6   4 | autocast                                                                                                                                                                                                                                                         
⠏ |          0 | autograd                  

libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
[1]    3743 abort      R --no-save
```

dfalbel:
Unfortunatelly, this is a limitation of `pkgload` when trying the package twice in the same session it never get's to fully unload the previous version and causes incompatibility issues, and symbol clashes.

sebffischer:
Ah thx! I can add this information to the contributing guide

PR #1219, opened 2024-12-11
docs(contributing): add information about segfaults

sebffischer:
Solves #1218


dfalbel:
Just for more info, the problem should not be running the tests twice, but using `devtools::test()` which causes `devtools::load_all()` to run and then causing the issue that you see.

PR #1220, opened 2024-12-11
fix(build): building lantern now works correctly

sebffischer:


dfalbel:
Thanks! You shouldn't need this script if you set `Sys.setenv(BUILD_LANTERN=1)`.

PR #1227, opened 2025-01-03
Avoid non-API calls

shikokuchuo:
Solves the CRAN check NOTEs for `PRCODE`, `PRENV` and `Rf_findVarInFrame3`.

As the fundamental R types are defined as a union at the top of [Defn.h](https://github.com/wch/r-source/blob/446d3f76137186bf63de36776a13f171cb0579a3/src/include/Defn.h#L197), we're able to simply use alternative accessors.


dfalbel:
Thanks @shikokuchuo !

Issue #1233, opened 2025-01-15
Scalar Torch Tensors cannot be loaded after serialization


sebffischer:
``` r
library(torch)
torch_load(torch_serialize(torch_tensor(1)))
#> torch_tensor
#>  1
#> [ CPUFloatType{1} ]
torch_load(torch_serialize(torch_scalar_tensor(1)))
#> Error in cpp_tensor_from_buffer(buffer, shape, list(dtype = dtype)): Not compatible with requested type: [type=list; target=double].
```

<sup>Created on 2025-01-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1235, opened 2025-01-15
fix(ignite): making state saveable, param_groups modifieable

sebffischer:
It was impossible to save the state because of:

* https://github.com/mlverse/torch/issues/1233
* undefined tensors were part of the state such as max_exp_avg_sq for adam with amsgrad = FALSE. We now keep them as 0-sized tensors as undefined tensors are not serializeable. (The reason we keep them at all is that it simplifies the saving and loading of state dicts easier)

This PR also improves the tests by removing an unnecessary call to `torch_manual_seed()` that made the tests deterministic

### DIFF
--- a/R/ignite.R
+++ b/R/ignite.R
@@ -117,7 +117,8 @@ OptimizerIgnite <- R6::R6Class(
     #' The parameter groups of the optimizer.
     param_groups = function(rhs) {
       if (!missing(rhs)) {
-        prev_param_groups <- self$state_dict()$param_groups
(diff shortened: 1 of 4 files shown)


sebffischer:
@dfalbel Can you tag this with 'lantern' again please? I forgot to test the saving of the state dict which was not possible.

dfalbel:
Done! @sebffischer

sebffischer:
@dfalbel this can be reviewed :)

Issue #1252, opened 2025-01-22
Trace-Jitting fails when combined with jit-compiled module


sebffischer:
Below, the input type of the trace-jitted module is inferred as `Tensor` even though it received a `list[Tensor]` during tracing.

``` r
library(torch)
sum_jit = jit_compile("
def sum_jit(xs: List[Tensor]):
  output = torch.zeros_like(xs[0])
  for x in xs:
    output = output + x
  return output
")$sum_jit


net2 = nn_module("nn_mix",
  initialize = function() {
    self$linear = nn_linear(10, 1)
  },
  forward = function(x) {
    self$linear(sum_jit(x))
  }
)()

net2_jit = jit_trace(net2, list(torch_randn(10, 10), torch_randn(10, 10)))

net2_jit(list(torch_tensor(1), torch_tensor(2), torch_tensor(3)))
#> Error in cpp_call_jit_script(ptr, inputs): forward() Expected a value of type 'Tensor (inferred)' for argument 'x' but instead found type 'List[Tensor]'.
#> Inferred 'x' to be of type 'Tensor' because it was not annotated with an explicit type.
#> Position: 1
#> Declaration: forward(__torch__.nn_mix self, Tensor x) -> Tensor
#> Exception raised from checkArg at /Users/runner/work/libtorch-mac-m1/libtorch-mac-m1/pytorch/aten/src/ATen/core/function_schema_inl.h:20 (most recent call first):
#> frame #0: c10::Error::Error(c10::SourceLocation, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) + 52 (0x10375011c in libc10.dylib)
#> frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 140 (0x10374cd6c in libc10.dylib)
#> frame #2: void c10::FunctionSchema::checkArg<c10::Type>(c10::IValue const&, c10::Argument const&, std::__1::optional<unsigned long>) const + 312 (0x14b376d74 in libtorch_cpu.dylib)
[...]

PR #1253, opened 2025-01-23
fix(tracing): respect mode in R forward function

sebffischer:
I realized that because TorchScript is statically typed, we would need to infer the types for the forward method that selects trainforward or evalforward depending on the mode (if we want to register this method in the C++ ScriptModule).
While I still think this is the cleaner approach, I don't immediately now how to do this, so I would defer this to a later point. For now this also works I think.

Resolves Issue #1252

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -15345,10 +15345,6 @@ cpp_jit_script_module_add_method <- function(self, method) {
     invisible(.Call(`_torch_cpp_jit_script_module_add_method`, self, method))
 }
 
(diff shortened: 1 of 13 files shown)


sebffischer:
@dfalbel once again I need the 'lantern' tag

dfalbel:
Just added the lantern tag. I also gave you some additional roles in the repository, so you should be able to mark PR labels now.

sebffischer:
Tests seem to be passing @dfalbel

PR #1255, opened 2025-01-23
feat(printer): improve printer for optimizer generator

sebffischer:


dfalbel (commented):
Can you show an example of an `optimizer_generator`?

sebffischer:
```r
> torch::optim_adamw
function (params, lr = 0.001, betas = c(0.9, 0.999), eps = 1e-08, 
    weight_decay = 0.01, amsgrad = FALSE) 
{
    d$new(params = params, lr = lr, betas = betas, eps = eps, 
        weight_decay = weight_decay, amsgrad = amsgrad)
}
<bytecode: 0x10a0ee210>
<environment: 0x10a188168>
attr(,"Optimizer")
<optim_adamw> object generator
  Inherits from: <inherit>
  Public:
    initialize: function (params, lr = 0.001, betas = c(0.9, 0.999), eps = 1e-08, 
    loop_fun: function (group, param, g, p) 
    step: function (closure = NULL) 
    clone: function (deep = FALSE) 
  Parent env: <environment: 0x10a7e40e0>
  Locked objects: FALSE
  Locked class: FALSE
  Portable: TRUE
attr(,"class")
[1] "optim_adamw"               "torch_optimizer_generator"
>
```

The PR ensures that we now just print the module and not the function anymore, just like it's done for `nn_module`:
https://github.com/mlverse/torch/blob/eb72cce6dadfef45a76ec56630f4e0c8609a507d/R/nn.R#L653

dfalbel:
Cool, nice!

PR #1258, opened 2025-01-26
Ready for blackwell and ARM runners

johnnynunez:


dfalbel:
This won't really work, because it requires changing the LibTorch version we support too.

Issue #1260, opened 2025-01-28
Ignite optimizers are incompatible with LR schedulers


sebffischer:
This is because currently the ignite optimizers don't allow setting a state that does not exist in C++.
I will open a PR soon.

PR #1261, opened 2025-01-28
fix(ignite): allow to set param_groups to arbitrary values

sebffischer:
This is needed for compatability with LR schedulers

Resolves Issue #1260


dfalbel (approved):
LGTM. I'll wait for tests and then merge

sebffischer:
It's only the usual failures.

PR #1263, opened 2025-01-29
Avoid unlock binding

dfalbel:
CRAN adds this note:

```
Found the following possibly unsafe calls:
File ‘torch/R/script_module.R’:
  unlockBinding("forward", self)
```

Which I think it's better to avoid for submission,


R/script_module.R
```diff
@@ -171,18 +181,7 @@ nn_ScriptModule <- R6::R6Class(
     },
     respects_mode = FALSE,
     update_forward_to_respect_mode = function() {
```
sebffischer, line 183:
```suggestion
    respect_mode = function() {
```

sebffischer (changes requested):

sebffischer (approved):

PR #1266, opened 2025-01-30
Align FR translation to 0.14.0

cregouby:


dfalbel:
Thanks @cregouby ! What's the process for generating those? Should I always do it before release?

cregouby:
Hello @dfalbel 

Thanks for merging my (late) contribution.

the complete process is in here : https://michaelchirico.github.io/potools/articles/developers.html
the TL;DR process, for update only, is

``` r
#|label: developer extracts messages
library(potools)
# update message translation file
po_update(here::here()))
```

This updates the common (english) `po/R-torch.pot` file.

Now each translator is ready to work to update its own `po/R-__LANG__.po` (clear text) translation file, and then to compile it into the target folder with `po_compile(here::here())`

About the release process, I would add two translation tasks
- [ ] update the `po/R-torch.pot` message file
- [ ] inform the translators and collect back updated translations

dfalbel:
Thanks, I'll make sure to do that for the next release. I'll make a patch release of 0.14 soon too, to include those changes.

cregouby:
I may propose a P.R. to `usethis::use_release_isssue()`...

PR #1285, opened 2025-02-26
Update installation.Rmd - fix CUDA version

statist-bhfz:
Maybe other sections should also be updated


dfalbel:
Thanks @statist-bhfz !

PR #1302, opened 2025-03-25
Add CUDA memory snapshot to help debug CUDA memory use

CoreyGiles:
Implemented bindings for LibTorch `_record_memory_history` and `_memory_snapshot_pickled`, similar to the python implementation: [https://pytorch.org/docs/stable/torch_cuda_memory.html](https://pytorch.org/docs/stable/torch_cuda_memory.html).

The created memory snapshots can be uploaded to [https://pytorch.org/memory_viz](https://pytorch.org/memory_viz) and used to debug CUDA OOM errors, see blog posts: [https://pytorch.org/blog/understanding-gpu-memory-1/](https://pytorch.org/blog/understanding-gpu-memory-1/) and [https://pytorch.org/blog/understanding-gpu-memory-2/](https://pytorch.org/blog/understanding-gpu-memory-2/).

Included minor details in the memory management vignette, which seemed like a logical place to outline the feature.


src/lantern/src/Cuda.cpp
```diff
@@ -234,4 +235,26 @@ void* _lantern_cuda_get_rng_state (int device) {
   auto gen = at::detail::getCUDAHooks().getDefaultCUDAGenerator(device);
   return make_raw::Tensor(gen.get_state());
   LANTERN_FUNCTION_END
+}
+
+void _lantern_cuda_record_memory_history(const std::string* enabled, const std::string* context, const std::string& stacks, size_t max_entries) {
```
dfalbel, line 240: Unfortunatelly we can't have C++ strings across the liblantern and R boundaries (because they might be compiled with different non ABI compatible compilers). You need to take them as `void *` arguments and convert them to proper strings using `from_raw::string()`. See eg: for an example:

https://github.com/mlverse/torch/blob/ca61df853546f6e2248f5de718d0f9d9c7163a8c/src/lantern/src/Compile.cpp#L28-L33

dfalbel (changes requested):
Thanks @CoreyGiles ! The PR looks great!

I've made a comment. Unfortunately, lantern can only export functions that take simple C data types as arguments and only return simple C data types too.

dfalbel (approved):

PR #1303, opened 2025-03-26
feat: allow to pass R vector as loss weights

sebffischer:
Just a very minor convenience improvement

### DIFF
--- a/R/nn-loss.R
+++ b/R/nn-loss.R
@@ -13,7 +13,7 @@ nn_weighted_loss <- nn_module(
   inherit = nn_loss,
   initialize = function(weight = NULL, reduction = "mean") {
     super$initialize(reduction)
-    self$register_buffer("weight", weight)
+    self$register_buffer("weight", torch_tensor(weight))
   }
 )
 


R/nn-loss.R
```diff
@@ -13,7 +13,7 @@ nn_weighted_loss <- nn_module(
   inherit = nn_loss,
   initialize = function(weight = NULL, reduction = "mean") {
     super$initialize(reduction)
-    self$register_buffer("weight", weight)
+    self$register_buffer("weight", if (!is.null(weight)) torch_tensor(weight))
```
dfalbel, line 16: I'm not sure, if someone would pass `weight` and expect it to be something that they can modify in-place. But it seems `torch_tensor()` is creating a copy of the tensor:

``` r
library(torch)
x <- torch_tensor(runif(10))
y <- torch_tensor(x)
y$add_(x)
#> torch_tensor
#>  1.3458
#>  1.9408
#>  1.6358
#>  1.8803
#>  0.0525
#>  0.5098
#>  1.4402
#>  0.9083
#>  1.8566
#>  1.8377
#> [ CPUFloatType{10} ]
torch_allclose(x, y)
#> [1] FALSE
```
 We might want to only call `torch_tensor()` if `weight` is not a tensor.

dfalbel (changes requested):

R/nn-loss.R
sebffischer: thanks, updated it!

PR #1308, opened 2025-04-22
fix(website): render math correctly

### LINKED ISSUE #1289: docs: Documentation does not render appropriately
The equations for the loss function do not render appropriately: 

https://github.com/mlverse/torch/blob/7ff1cdd49854569d227b53dab8dd410ff4564dc0/R/nn-loss.R#L28

Image

Additionally, it is not clear how to use this function. It would be great if there was a minimal example in the docs as well.

sebffischer:
Fixes #1289


sebffischer:
The test failures are not related to this change I think.

Issue #1209, opened 2024-11-06
Feature request for `nn_transformer_encoder`, `nn_transformer_encoder_layer`, `nn_transformer_decoder`, and `nn_transformer_decoder_layer`


MaximilianPi:
It would be nice to have the transformer objects in Torch. 

Actually, I have already implemented most of them (by translating the pytorch code https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py ), so I could do a PR if you want.

cregouby:
Please do !

PR #1313, opened 2025-05-18
feat: nn-transformer module

OJWatson:
Hi everyone,

I've had a go at adding the `TransformerEncoderLayer` and `TransformerEncoder` modules from PyTorch, hoping to eventually address issue #1209. I'm opening this as a draft PR because the R torch codebase is really new to me, and I suspect there are many aspects of my implementation that are wrong, could be improved or adjusted, and I am hoping that this may be helpful for others to help with as well (if not please ignore). 

In particular, I'd really appreciate guidance on the following:

- C++: I am just a bit lost here about how to start to bring across the cpp for this (similar to https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/transformer.cpp?) into the R torch (tried looking at rnn to get my bearings but struggling) so I have not tried yet. 

- Testing: Are the provided tests helpful and sufficient, or is there anything critical that's missing? I'm not sure if I've covered all edge cases or if additional tests would be beneficial but have checked coverage is 100% against the new `nn-transfer.R`

- Code structure: I've tried to follow existing conventions, which I based on comparing how `nn-rnn` module was developed in relation to PyTorch equivalent but I'd appreciate any pointers if things are structured incorrectly or could be more idiomatic.

- Argument checks and defaults: Have I included enough checks on the provided arguments? Are the defaults sensible, or is anything critical missing? (PyTorch has significantly more but was basing on `nn-rnn`). 

My plan, once this initial PR is in better shape, is to add the `TransformerDecoderLayer` and then the full `Transformer` module, and then if I can figure it out or if someone else can help with adding the cpp for this. Before proceeding further, it would be really helpful to have your feedback and pointers to see if I’m at all on the right track.

Thanks very much in advance and hopefully this is helpful. 

All the best,

OJ


dfalbel:
Hi @OJWatson,

Thank you very much for your PR. I'm happy to merge as is as it's an awesome addition to the API. We can take a look at adding the C++ part, but in general we prefer using the same codepaths as in Python instead of using the Aten functions that only get used by the C++ backend.

Issue #1320, opened 2025-05-28
Possibly exploding object size when defining a `nn_module` in the `$initialize()` method.


sebffischer:
Because the default of `parent_env` for `nn_module` is `parent.frame()`, defining an `nn_module` in an environment that is not a special environment (such as the global environment or a package environment) can lead to exploding object sizes when serializing.

The special environments (global, pkg env) are no problem, because their serialization is handled special anyway.

``` r
library(torch)

nn_module1 = nn_module("nn_module1",
  initialize = function(data) {
    data
    self$layer = nn_module("layer",
      initialize = function() {
        self$linear = nn_linear(10, 1)
      },
      forward = function(x) {
        self$linear(x)
      }
    )()
  },
  forward = function(x) {
    self$layer(x)
  }
)

pryr::object_size(nn_module1(rnorm(1L)))
#> 499.06 kB
pryr::object_size(nn_module1(rnorm(10000000L)))
#> 80.50 MB


net = nn_module1(rnorm(1))
net$layer$forward |> environment() |> parent.env() |> parent.env() |> with(data)
#> [1] 1.308416
```

<sup>Created on 2025-05-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I also think this is not a crazy edge case, but something that can actually happen and some big object (like training data) can cause issues here.

I am not really sure what to do about this.

PR #1322, opened 2025-05-28
Possible fix for exploding object size

sebffischer:
I am not yet sure whether this has any negative consequences that are problematic.

What do you think about this approach?

If this is a desirable solution to #1320 then I would still need to add more tests.

``` r
devtools::load_all("~/gh/torch")
#> ℹ Loading torch

nn_module1 = nn_module("nn_module1",
  initialize = function(data) {
    self$layer = nn_module("layer",
      initialize = function() {
        self$linear = nn_linear(10, 1)
      },
      forward = function(x) {
        self$linear(x)
      }
    )()
  },
  forward = function(x) {
    self$layer(x)
  }
)

pryr::object_size(nn_module1(rnorm(1L)))
#> 916.75 kB
pryr::object_size(nn_module1(rnorm(1000000L)))
#> 916.75 kB


# What will not work:

nn_module2 = nn_module("nn_module1",
  initialize = function(data) {
    f = nnf_relu
    self$layer = nn_module("layer",
      initialize = function() NULL,
      forward = function(x) {
        f(x)
      }
    )()
  },
  forward = function(x) {
    self$layer(x)
  }
)

nn_module2()(torch_randn(1))
#> Error in f(x): could not find function "f"
```

<sup>Created on 2025-05-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/nn.R
+++ b/R/nn.R
@@ -1,6 +1,16 @@
 #' @include utils-data.R
 NULL
 
+default_parent_env = function() {
+  env = parent.frame(2)
+  pe = parent.env(env)
+  if (all(sapply(c(".__active__", "self", "private", "super"), function(field) exists(field, pe)))) {
+    get("inherit", pe)$parent_env
+  } else {
+    env
+  }
+}
+
 get_inherited_classes <- function(inherit) {
   inherit_class <- inherit$public_fields$.classes
   # Filter out classes that we eventually add in our normal flow.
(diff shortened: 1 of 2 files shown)


dfalbel (commented):
Looks reasonable, although I do think we should not recommend people to define modules within modules. It's better to define them in top level environments. In the worst case scenario its always possible to create a generic module that takes a closure to initialize parameters a a `forward`closure. Eg: 

```
nn_generic_module <- nn_mopdule(
   inititialize = function(init, fwd) {
      self$fwd <- fwd
      self$pars <- lapply(init(), nn_parameter) 
   },
   forward = function(x) {
      self$fwd(pars, x)
   } 
)
```

PR #1328, opened 2025-06-23
[0.15.0] update fr translation

cregouby:
Hello, 

Here is the updated fr translation for 0.15.0


R/install.R
```diff
@@ -543,7 +543,7 @@ installer_message <- function(msg) {
   if (!is_truthy(Sys.getenv("TORCH_INSTALL_DEBUG", FALSE)))
     return(invisible(msg))
   names(msg) <- rep("i", length(msg))
-  cli_inform(msg, class = "torch_install", env = parent.frame())
+  cli_inform(msg, env = parent.frame())
```
dfalbel, line 546: We may want to add a `class` argument to our `cli_inform` wrapper so we don't miss it. I don't remember why this was added, but I think this can be useful for capturing the message.

dfalbel (approved):
Thanks @cregouby

PR #1409, opened 2026-02-05
Bump to libtorch 2.8.0

TroyHernandez:
## Summary

- Update libtorch from 2.7.1 to 2.8.0
- Generate Declarations-2.8.0.yaml from PyTorch v2.8.0 source
- Fix Linux download URL format (2.8.0 dropped `cxx11-abi` from filenames)
- Update lantern CMakeLists.txt to match

## Why 2.8.0?

PyTorch 2.8 introduced the **stable libtorch ABI**, which means code compiled against 2.8 can link with 2.9, 2.10, etc. without recompiling. This is a good foundation for future upgrades.

## API Changes

No breaking changes. Only 3 new functions added:
- `_fused_rms_norm`
- `_weight_int4pack_mm_with_scales_and_zeros`
- `_grouped_mm`

## Testing

Built and tested locally on Ubuntu 24.04 with CUDA 12.8. Basic tensor operations and matmul chains work correctly.

## Notes

This PR includes the code changes but lantern binaries would need to be rebuilt against libtorch 2.8.0 and uploaded to the CDN.

Happy to help with any adjustments needed.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/man/torch_dtype.Rd
+++ b/man/torch_dtype.Rd
@@ -15,6 +15,8 @@
 \alias{torch_float16}
 \alias{torch_half}
 \alias{torch_bfloat16}
+\alias{torch_float8_e4m3fn}
+\alias{torch_float8_e5m2}
 \alias{torch_uint8}
 \alias{torch_int8}
 \alias{torch_int16}
@@ -54,6 +56,10 @@ torch_float16()
 torch_half()
 
 torch_bfloat16()
+
+torch_float8_e4m3fn()
+
+torch_float8_e5m2()
 
 torch_uint8()
 


TroyHernandez:
The `mlverse/libtorch-mac-m1` repo only has builds up to 2.7.1 — there's no `libtorch-x86_64-v2.8.0.zip` or `libtorch-arm64-v2.8.0.zip` yet. The macOS lantern build (`src/lantern/CMakeLists.txt:38`) downloads from there, so macOS CI will fail until those are uploaded.

PR #1451, opened 2026-04-22
Replace R_RunPendingFinalizers with weak reference flag check

dfalbel:
## Summary

- Replace `R_RunPendingFinalizers()` + `Rf_duplicate()` in `operator_sexp_tensor` with R's weak reference API
- Store an `R_MakeWeakRefC` weak reference in TensorImpl's pyobj slot instead of the raw SEXP
- Check the `READY_TO_FINALIZE` flag (bit 0 of `gp`, set by R's GC in `CheckFinalizers()` before finalizers run) to detect stale keys without forcing all pending finalizers to execute

### Why

`R_RunPendingFinalizers()` is called on **every** tensor return that hits the cached SEXP path (primarily in-place operations like `mul_`, `add_`, etc.). It forces execution of all pending finalizers across the entire R session, which is unnecessary work. This is particularly costly in tight loops like optimizer steps.

### How it works

R's GC processes weak references in two phases:
1. `CheckFinalizers()` — sets `READY_TO_FINALIZE` flag on weak refs whose keys are unreachable
2. `RunFinalizers()` — clears keys/values and runs finalizer callbacks

By checking the flag (via `LEVELS(weakref) & 1`), we can detect that a key has been scheduled for collection **before** the finalizer runs — closing the race window that `R_RunPendingFinalizers()` was solving.

## Test plan

- [x] Basic tensor operations (in-place and new-tensor) work correctly
- [x] GC stress test (500 iterations with `gc()` between ops) passes
- [x] Neural network forward/backward passes
- [x] Optimizer step works
- [ ] Full test suite (`devtools::test()`) — running
- [ ] gctorture tests

### DIFF
--- a/src/torch_api.cpp
+++ b/src/torch_api.cpp
@@ -5,11 +5,23 @@
 // torch's C API. Most of functions here should not be directly used, but they
 // are called by torch type wrappers.
 
-static inline void tensor_finalizer(SEXP ptr) {
-  auto xptr = Rcpp::as<Rcpp::XPtr<XPtrTorchTensor>>(ptr);
-  // When the the xptr is released, we might call this function with an invalid ptr.
+// READY_TO_FINALIZE is bit 0 of the gp field. R's GC sets this flag in
+// CheckFinalizers() BEFORE running the actual weak-ref finalizer, so we can
+// detect that a key has been scheduled for collection without calling
+// R_RunPendingFinalizers().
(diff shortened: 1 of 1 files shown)


dfalbel:
Closing: benchmarks showed the weak ref approach doesn't improve performance over the current R_RunPendingFinalizers path.

PR #1470, opened 2026-07-16
Add experimental Lantern-backed Tensor API

dfalbel:
## What

Adds a `torch::experimental::Tensor` C++ facade backed by Lantern calls. It supports direct use in Rcpp exports, including fluent code such as `input.relu().reshape({-1})`.

The initial handwritten facade covers core Tensor behavior and conversions, while generated methods expose the broader Tensor namespace from `src/gen-namespace.cpp`. The generator currently emits 758 overloads and audits coverage of 815 public Lantern Tensor entry points.

## Why

Packages using torch currently receive handle types but cannot write a LibTorch-style Tensor API without linking directly against LibTorch. This facade keeps the existing Lantern ABI boundary while providing a familiar C++ interface.

## Impact

- Adds an opt-in API under `torch::experimental`; existing APIs are unchanged.
- Resolves Lantern symbols through registered R C-callables.
- Adds representative runtime tests for Tensor input/output, chaining, generated methods, integer arguments, and boolean returns.
- Adds a freshness and coverage test for generated methods.

## Validation

- `python3 tools/generate-experimental-tensor.py --check` — passes (758 generated overloads; 815 Lantern Tensor entry points covered).
- `testthat::test_file("tests/testthat/test-experimental-cpp-codegen.R")` — passes.
- `testthat::test_file("tests/testthat/test-experimental-cpp-api.R")` — passes all 5 runtime assertions against the installed branch and configured Lantern runtime.
- The package was installed from this branch with `BUILD_LANTERN=1 R CMD INSTALL .` before running the tests.


dfalbel:
Added the experimental tensor creation API in `e6ae16dc5`.

Exposed under `torch::experimental`:
- `empty`, `empty_strided`, `empty_like`
- `zeros`, `zeros_like`, `ones`, `ones_like`
- `rand`, `rand_like`, `randn`, `randn_like`
- `randint`, `randint_like`, `randperm`
- `arange`, `linspace`, `logspace`
- `eye`, `full`, `full_like`, `scalar_tensor`

Each factory supports LibTorch-style default options, and applicable overloads accept the existing `torch::TensorOptions` handle. Compile coverage exercises every operation. The focused suite passes all 10 runtime assertions, and the generator audit still passes with 758 generated Tensor methods and 815 Lantern Tensor entry points.

Issue #1463, opened 2026-06-06
Problems loading lantern.so when running R CMD check.


ralmond:
I'm working on a package which uses torch.  I installed torch by doing

```
install.package("cuda12.8",repos=...)
install.packge("torch")
torch::install_torch()
```
Which apparently ran without errors.

Things seem to be working properly when I run my testthat tests.  However, when I run
R CMD check to thoroughly test my package, I'm consistently getting the following error:

```
Warning message:
ℹ torch failed to start, restart your R session to try again.
ℹ You might need to reinstall torch using `install_torch()`
✖ /home/ralmond/R/x86_64-pc-linux-gnu-library/4.5/torch/lib/liblantern.so -
  libcudart.so.12: cannot open shared object file: No such file or directory
Caused by error:
! /home/ralmond/R/x86_64-pc-linux-gnu-library/4.5/torch/lib/liblantern.so - libcudart.so.12: cannot open shared object file: No such file or directory 
Error : Lantern is not loaded. Please use `install_torch()` to install additional dependencies.
Error: unable to load R code in package ‘CPTtorch’
Execution halted
ERROR: lazy loading failed for package ‘CPTtorch’
* removing ‘/home/ralmond/ralmond1/Projects/CPTtorch.Rcheck/CPTtorch’
```

Checking in my torch directory, I see:

```
$ ls libcudart*
libcudart-256e6409.so.12  libcudart-c3a75b33.so.12
```

Is there a missing simlink?

ralmond:
Manually adding symlinks to `system.file("lib",package="torch")`, for example adding
```
$ ln -s libcudart-c3a75b33.so.12 libcudart.so.12
```
and similar for other libs whi only have a `libXXX-XXXXXXXX.so.N` version, sort of works around the problem.  `R CMD check` will now load the code and the rest of the run works pretty normally.

However, I still get the following warning:

```
W  checking whether the package can be loaded with stated dependencies (1.7s)
   Loading required package: methods
   Loading required package: torch
   Torch libraries are installed but loading them was unsuccessful. Please
   reinstall torch with `install_torch(reinstall = TRUE)` You can disable this
   check by setting `TORCH_VERIFY_LOAD` to "FALSE" Torch libraries installed in:
   '/home/ralmond/R/x86_64-pc-linux-gnu-library/4.5/torch'. Error: Lantern is not
   loaded. Please use `install_torch()` to install additional dependencies. In
[...]

PR #1478, opened 2026-09-01
Guard the cudatoolkit lib_path() call during startup

TroyHernandez:
`load_cudatoolkit_libs()` calls the `cuda*` package's `lib_path()` unguarded, so any error raised there propagates out of `.onLoad` and is reported as torch failing to start.

On a machine with a `cuda*` package installed:

```
$ R_DEFAULT_PACKAGES=NULL Rscript --vanilla -e 'loadNamespace("torch")'
ℹ torch failed to start, restart your R session to try again.
ℹ You might need to reinstall torch using `install_torch()`
✖ could not find function "packageName"
```

The reported fault is torch; the actual fault is in the toolkit package. That particular trigger is mlverse/cudatoolkit#5, and `R CMD check` clears `R_DEFAULT_PACKAGES` for its whole driver process, so it fires there too.

Not hypothetical: the same signature is in [#1463 (comment, 2026-07-05)](https://github.com/mlverse/torch/issues/1463#issuecomment-4887670975), where it outlives the symlink workaround that thread is otherwise about. For a package that touches torch at load time the cost is not a NOTE, it is three `R CMD check` WARNINGs and `.onLoad failed in loadNamespace()`.

### The existing guard cannot fire

```r
lib_path <- getExportedValue(pkg_name, "lib_path")()
if (!dir.exists(lib_path)) return(invisible(FALSE))
```

`cuda*::lib_path()` is `system.file("lib", package = ..., mustWork = TRUE)`, which errors rather than returning `""` when the directory is missing, so control never arrives at `dir.exists()` with a bad path.

### Change

Catch the error, warn naming the toolkit package, and continue without preloading. This matches how the `dyn.load()` calls further down are already handled.

Verified against the real failing call:

```
# before
ERROR propagates: could not find function "packageName"

# after
Warning message:
ℹ Skipping CUDA library preload from cuda12.8.
✖ could not find function "packageName"
guard returned: NULL -> caller returns invisible(FALSE)
```

`cli::cli_warn()` was checked in the same `R_DEFAULT_PACKAGES=NULL` environment, so the handler does not reintroduce the failure it reports.

Startup then continues with whatever libraries remain available. Whether that lands on a working session depends on the build: a CPU build, or one with another usable CUDA source, starts; a CUDA build that needed the skipped libraries still fails, but names the toolkit package rather than torch.

### One question

Possibly also upstream of #1463's *original* symptom. `load_cudatoolkit_libs()` exists to `dyn.load()` the toolkit's CUDA libraries before `cpp_lantern_init()`, so if it aborts or silently no-ops, `liblantern.so` has nothing to resolve `libcudart.so.12` against. Not asserting it — the original trace there reaches `cpp_lantern_init()`, so the preload returned rather than errored in that invocation.


TroyHernandez:
The 4 R CMD check failures here are the "less than" block in `tests/testthat/test-distributions-constraints.R`, which this PR does not touch. #1482 fixes both and its check jobs are green. I'll merge main into this branch once #1482 is in so the checks rerun.

